PR #9, opened 2019-06-11
COnscreenTimer and COnscreenTimerEntry

ShFil119:
Taxi mission etc...


src/Timer.h
```diff
@@ -2,6 +2,7 @@
 
 class CTimer
 {
+public:
```
aap: Don't make this public again. Use the accessors.

src/User.cpp
```diff
@@ -2,9 +2,158 @@
...
-WRAPPER void COnscreenTimer::ProcessForDisplay(void) { EAXJMP(0x4292E0); }
+char *CTheScripts::ScriptSpace = (char*)0x74B248;
+
+int COnscreenTimer::Init() {
+    m_bDisabled = 0;
+    m_sEntries[0].m_nTimerOffset = 0;
+    m_sEntries[0].m_nCounterOffset = 0;
+
+    for(int i = 0; i < 10; i++) {
+        m_sEntries->m_aTimerText[i] = 0;
```
aap: m_sEntries[0]

src/audio/DMAudio.h
```diff
@@ -15,5 +15,6 @@ class cDMAudio
 	Char GetCDAudioDriveLetter(void);
 	Bool CheckForAnAudioFileOnCD(void);
 	void ChangeMusicMode(UInt8 mode);
+    char PlayFrontEndSound(unsigned int, unsigned int);
```
aap: I think this is `void`

src/control/Replay.h
```diff
@@ -4,4 +4,6 @@ class CReplay
 {
 public:
 	static void Display(void);
+
+    static char &Mode;
```
aap: better use (u)int8 here.

src/render/Hud.cpp
```diff
@@ -2,5 +2,7 @@
 #include "patcher.h"
 #include "Hud.h"
 
+char &CHud::m_Wants_To_Draw_Hud = *(char*)0x95CD89;
```
aap: make this a bool

aap:
I added some comments.
Also generally:
- indent with tabs!
- don't write `unsigned`, we have typedefs for uint32 and so on.
- also I'd like to avoid `auto`. GTA uses a simple subset of c++ and auto is hardly ever useful.
- m_sEntries is an array (even if it just has one element). Make a variable in config.h for the number of onscreen timers and write loops everywhere. I know it doesn't matter for the final result but I still think it's better that way.

PR #15, opened 2019-06-13
Fix value of CReplay mode

ShFil119:
It should have value 1. (enum starts from 0)
https://github.com/GTAmodding/re3/commit/01babd43b1e500b36d67d28077b5cffd13212068#diff-cdb9d94a821ead0296d92a18149ff5fbL36


aap:
Oh sorry, my fault :(

PR #26, opened 2019-06-20
Big CPed update

erorcun:
Signed-off-by: eray orçunus <erayorcunus@gmail.com>

### DIFF
--- a/src/General.h
+++ b/src/General.h
@@ -40,6 +40,48 @@ public:
 		}
 	}
 
+	static float LimitRadianAngle(float angle)
+	{
+		if (angle < -25.0f)
+			angle = -25.0f;
+
+		if (angle > 25.0f)
+			angle = 25.0f;
+
+		float result = angle;
+
+		while (result >= PI) {
+			result -= 2 * PI;
+		}
+
+		while (result < -PI) {
+			result += 2 * PI;
+		}
+	
+		return result;
+	}
+
+	static float GetRadianAngleBetweenPoints(float x1, float y1, float x2, float y2)
+	{
+		float x = x2 - x1;
+		float y = y2 - y1;
+
+		if (y == 0.0f)
+			y = 0.0001f;
+
+		if (x > 0.0f) {
+			if (y > 0.0f)
+				return 2 * PI - atan2(x / y, 1.0f);
+			else
+				return -atan2(x / y, 1.0f);
+		} else {
+			if (y > 0.0f)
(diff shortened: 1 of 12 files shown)


src/General.h
```diff
@@ -40,6 +40,48 @@ class CGeneral
 		}
 	}
 
+	static float LimitRadianAngle(float angle)
+	{
+		if (angle <= -25.0f)
```
aap: actually < in the code if you want to be super accurate

src/General.h
```diff
@@ -40,6 +40,48 @@ class CGeneral
...
+			angle = -25.0f;
+
+		if (angle >= 25.0f)
```
aap: same here, >

src/General.h
```diff
@@ -40,6 +40,48 @@ class CGeneral
...
+
+	static float GetRadianAngleBetweenPoints(float x1, float y1, float x2, float y2)
+	{
+		float x = x2 - x1;
+		float y = y2 - y1;
+
+		if (y == 0.0f)
+			y = 0.0001f;
+
+		if (x <= 0.0f) {
```
aap: hint: floats are almost never compared with <= and >=. So in this case the original code had > and < and the ifs/elses were flipped. Not a big deal but i think it's nicer to stick to the original code.

src/entities/Ped.cpp
```diff
@@ -176,6 +181,100 @@ static char WaitStateText[21][16] = {
...
+	bool foundDead = false;
+	bool foundOnTheFloor = false;
+	bool foundBelow = false;
+	float angleDiff;
+	float distance;
+
+	if (!CGame::nastyGame)
+		return NO_PED;
+
+	while (true) {
```
aap: This is a for loop: `for(currentPedId = 0; currentPedId < player->m_numNearPeds; currentPedId++)`

src/entities/Ped.h
```diff
@@ -2,17 +2,27 @@
...
+#include "AnimBlendAssociation.h"
+#include "WeaponInfo.h"
 
 struct CPathNode;
 
 enum {
 	PED_MAX_WEAPONS = 13
 };
 
+enum PedOnGroundState : int32 {
```
aap: I don't think we need int32 here

aap:
Looking quite good. Please change `type* var;` to `type *var;`. I saw you did it in a few places already.

PR #47, opened 2019-06-24
Fix values in PEDTYPE's enum

ShFil119:
Check `CPedType::FindPedType` to verify.


aap:
I don't see how this changes the values of the enum....

gennariarmando:
Last values have been changes, you might want to move up and sort them correctly, no need to add "= value" to the ones that already have that position.

aap:
pushed my own fix, thanks.

PR #50, opened 2019-06-25
Fix car animations & M16 bug

erorcun:
Signed-off-by: eray orçunus <erayorcunus@gmail.com>


gennariarmando:
Line 1453:
What is that "GetPosition() = neededPos" used for?
You might want to change it to SetPosition(neededPos).
Line 1434:
Might also change this "* 0.0016666667f" to "/ 600.0f"

PR #59, opened 2019-06-26
Added Gangs

Nick007J:


src/control/Gangs.cpp
```diff
@@ -0,0 +1,71 @@
...
+void CGangs::SaveAllGangData(uint8 *buffer, uint32 *size)
+{
+	tGangSaveData *data = (tGangSaveData*)buffer;
+	data->tag[0] = 'G';
+	data->tag[1] = 'N';
+	data->tag[2] = 'G';
+	data->tag[3] = '\0';
+	data->size = *size = sizeof(tGangSaveData);
+	for (int i = 0; i < TOTAL_GANGS; i++)
+		data->gangs[i] = *GetGangInfo(i);
```
aap: I'm not a fan on relying on the struct layout here. Can you do it like I did PedType

PR #67, opened 2019-06-28
Some Hud.cpp fixes.

gennariarmando:


src/render/Hud.cpp
```diff
@@ -205,7 +205,7 @@ void CHud::Draw()
 				}
 				else {
 					if (Mode == CCam::MODE_M16FIRSTPERSON_34 ||
-					    Mode == CCam::MODE_FIRSTPERSONPEDONPC_41 ||
+					    Mode == CCam::MODE_FIRSTPERSONPEDONPC_41 &&
```
aap: No, || was right here, but the != should be a ==. sorry for the bug

PR #77, opened 2019-06-30
More replay stuff + destructors for CPlaceable and up

Nick007J:


src/DamageManager.cpp
```diff
@@ -1 +1,21 @@
-#include "DamageManager.h"
\ No newline at end of file
+#include "common.h"
+#include "patcher.h"
+#include "DamageManager.h"
+
+#if 0
+WRAPPER void CDamageManager::SetDoorStatus(int door, unsigned int status) { EAXJMP(0x545920); }
+#else
+void CDamageManager::SetDoorStatus(int door, unsigned int status)
```
ShFil119: nit: using in32/uin32 as args would make code more multiplatform.

PR #82, opened 2019-06-30
More CPed

erorcun:


src/entities/CivilianPed.cpp
```diff
@@ -2,13 +2,18 @@
 #include "patcher.h"
 #include "CivilianPed.h"
 
-CCivilianPed::CCivilianPed(int pedtype, int mi)
+WRAPPER void CCivilianPed::ProcessControl(void) { EAXJMP(0x4BFFE0); }
+
+CCivilianPed::CCivilianPed(int pedtype, int mi) : CPed(pedtype)
```
ShFil119: nit: int32 - more multiplatform

erorcun: It was Nick's choice. I will write/ask to the group once it's merged for what we should do about it.

PR #87, opened 2019-07-02
Readme clean up

ShFil119:
I'd move rest of stuff into wiki.

To see how it looks:
https://github.com/ShFil119/re3/tree/readmeCleanUp


aap:
I'm not so sure yet. I don't know if i want things hidden in the wiki....i'll think about it

ShFil119:
Readme would point to wiki. So I think it's no a problem.

PR #88, opened 2019-07-02
Implemented frontend inputs.

### LINKED ISSUE #66: The "MP3 Player" icon is always present in audio settings.
Even when there's no audio files in the folder.
gta3_TiFI22bcDD

gennariarmando:
Fixed #66 and few other stuff.


src/Pad.h
```diff
@@ -1,5 +1,22 @@
 #pragma once
 
+// Some macros for frontend stuff.
+#define PAD_START (CPad::GetPad(0)->NewState.Start && !CPad::GetPad(0)->NewState.Start)
```
ShFil119: nit: imho it'd be nicer with lambdas.

aap: nah, no lambdas. we're writing old c++

aap: But still. get rid of all this and use the functions we have already (add the ones we don't)

PR #91, opened 2019-07-02
Reimplemented "correctly" frontend inputs.

gennariarmando:


aap:
Please don't open gta3.set in binary mode. I know it's stupid but that's how the game does it.

PR #108, opened 2019-07-03
Implement smallest functions from AudioManager

ShFil119:
I think in two more PR I should cover all AudioManager.


src/entities/Vehicle.h
```diff
@@ -24,6 +24,80 @@ enum eCarLock : uint8 {
 	CARLOCK_SKIP_SHUT_DOORS
 };
 
+enum eVehicleModel : int32
```
aap: what's this?

aap:
Ok, i'm pulling this now, but i'm not quite happy with the vehicle enum yet.

PR #153, opened 2019-07-20
premake and appveyor update

ThirteenAG:
- cleaned up configurations
- added .gitignore
- fixed compilation error without installed mfc
- copy postbuild event doesn't fail a build and only copies output file if it exists in second directory


.appveyor.yml
```diff
@@ -1,20 +1,46 @@
...
 environment:
   APPVEYOR_SAVE_CACHE_ON_ERROR: true
 install:
-  - set PATH=%APPVEYOR_BUILD_FOLDER%/bin;%PATH%
-before_build:
-  - mkdir "%APPVEYOR_BUILD_FOLDER%/build"
-  - cd "%APPVEYOR_BUILD_FOLDER%"
-  - premake5.exe vs2015
+- cmd: >-
+    git submodule update --init --recursive
```
ShFil119: Is it needed?

.appveyor.yml
```diff
@@ -1,20 +1,46 @@
...
+    premake-vs2019.cmd
 build:
-  project: c:\projects\re3\build\re3.sln
+  project: build/re3.sln
   verbosity: minimal
-artifacts:
-  path: bin/%CONFIGURATION%/re3.dll
+after_build:
+- ps: >-
+    $releases = "https://github.com/ThirteenAG/Ultimate-ASI-Loader/releases"
```
ShFil119: I'd add comment what's happening here. Also remove unneeded empty lines below.

.appveyor.yml
ThirteenAG: Not at the moment, no, but I figured it can't hurt to have it, in case dxsdk/rwsdk will be moved to submodule or something.

ShFil119:
Also maybe you should update binary of premake.

ThirteenAG:
>Also maybe you should update binary of premake.

Did that, seems like this repo already contains the latest version to date.

.appveyor.yml
ThirteenAG: That's how appveyor exports this file, I don't add empty lines specifically.

PR #212, opened 2019-09-28
CCutsceneMgr

Sergeanur:


src/core/CutsceneMgr.cpp
```diff
@@ -1,8 +1,427 @@
...
+CCutsceneMgr::Initialise(void)
+{
+	ms_numCutsceneObjs = 0;
+	ms_loaded = false;
+	ms_running = false;
+	ms_animLoaded = false;
+	ms_cutsceneProcessing = false;
+	ms_useLodMultiplier = false;
+
+	ms_pCutsceneDir = new CDirectory(512);
```
ShFil119: nit: maybe extracting `512` as enum, macro or variable?

src/core/CutsceneMgr.cpp
```diff
@@ -1,8 +1,427 @@
...
+	pPlayerPed->m_fCurrentStamina = pPlayerPed->m_fMaxStamina;
+	CPad::GetPad(0)->DisablePlayerControls |= PLAYERCONTROL_DISABLED_80;
+	CWorld::Players[CWorld::PlayerInFocus].MakePlayerSafe(true);
+}
+
+void
+CCutsceneMgr::SetHeadAnim(const char *animName, CObject *pObject)
+{
+	CCutsceneHead *pCutsceneHead = (CCutsceneHead*)pObject;
+	char szAnim[16];
```
ShFil119: also?

Sergeanur: How about I add `CUTSCENEDIRSIZE = 512` to Config enum and macro `#define CUTSCENENAMESIZE 8` to CutsceneMgr.h, so it would be
`ms_cutsceneName[CUTSCENENAMESIZE]`
and this string
`char szAnim[CUTSCENENAMESIZE*2]`?

ShFil119: sounds good

PR #231, opened 2019-10-10
Full CRestart

Sergeanur:


src/control/Restart.cpp
```diff
@@ -1,14 +1,262 @@
...
+	OverridePosition = CVector(0.0f, 0.0f, 0.0f);
+	bOverrideRestart = false;
+	NumberOfPoliceRestarts = 0;
+	NumberOfHospitalRestarts = 0;
+
+	for (int i = 0; i < NUM_RESTART_POINTS; i++) {
+		HospitalRestartPoints[i] = CVector(0.0f, 0.0f, 0.0f);
+		HospitalRestartHeadings[0] = 0.0f;
+		PoliceRestartPoints[i] = CVector(0.0f, 0.0f, 0.0f);
+		PoliceRestartHeadings[0] = 0.0f;
```
ShFil119: `HospitalRestartHeadings[i] = 0.0f;` and `PoliceRestartHeadings[i] = 0.0f;`?

src/control/Restart.cpp
```diff
@@ -1,14 +1,262 @@
...
+		+ sizeof(NumberOfPoliceRestarts)
+		+ sizeof(bOverrideRestart)
+		+ sizeof(uint8)
+		+ sizeof(uint16)
+		+ sizeof(OverridePosition)
+		+ sizeof(OverrideHeading)
+		+ sizeof(bFadeInAfterNextDeath)
+		+ sizeof(bFadeInAfterNextArrest)
+		+ sizeof(OverrideHospitalLevel)
+		+ sizeof(OverridePoliceStationLevel); // == 292
```
ShFil119: `static_assert` here?

Sergeanur: I used to have it there actually, but removed last minute. VALIDATESAVEBUF should do the trick.

PR #238, opened 2019-10-12
Audio8

ShFil119:
More more audio.


src/audio/AudioManager.cpp
```diff
@@ -3427,7 +3981,7 @@ cAudioManager::ProcessBoatMovingOverWater(cVehicleParams *params)
 	if(velocityChange <= 0.0005f && params->m_pVehicle->GetPosition().y) return 1;
 
 	velocityChange = min(0.75f, velocityChange);
-	multiplier = (velocityChange - 0.0005f) * 1.3342f;
+	multiplier = (velocityChange - 0.0005f) * 4 / 3;
```
Nick007J: Well, not sure if that "rounding" is good, maybe there was a reason for that weird float

src/audio/AudioManager.cpp
```diff
@@ -7802,18 +8774,66 @@ cAudioManager::ReleaseDigitalHandle() const
...
+cAudioManager::ReportCollision(CEntity *entity1, CEntity *entity2, uint8 surface1, uint8 surface2,
+                               float collisionPower, float velocity)
 {
-	EAXJMP(0x568410);
+	float dist;
+	CVector v1;
+	CVector v2;
+
+	if(m_bIsInitialised && m_nCollisionEntity >= 0 && !m_bUserPause &&
+	   (velocity >= 0.0016f || collisionPower >= 0.01f)) {
```
Nick007J: this is squared velocity, which might be worth noting in parameter name, so it may make sense to replace 0.0016f with SQR(0.04f). Also most likely this whole condition was inverted:
`if (!m_bIsInitialised || ... ) return`
would also explain >=s for float, which aap doesn't like :P

src/audio/AudioManager.cpp
```diff
@@ -7802,18 +8774,66 @@ cAudioManager::ReleaseDigitalHandle() const
...
+		if((entity1->m_status & 7) == ENTITY_TYPE_BUILDING) {
```
Nick007J: `if (entity1->IsBuilding()`. Moreover, this is incorrect at all I think, you need to compare `entity1->m_type`.

src/audio/AudioManager.cpp
```diff
@@ -7802,18 +8774,66 @@ cAudioManager::ReleaseDigitalHandle() const
...
+			v1 = v2 = entity2->GetPosition();
+		} else if((entity2->m_status & 7) == ENTITY_TYPE_BUILDING) {
+			v1 = v2 = entity1->GetPosition();
+		} else {
+			v1 = entity1->GetPosition();
+			v2 = entity2->GetPosition();
+		}
+		CVector pos = (v1 + v2) * 0.5f;
+		dist = GetDistanceSquared(&pos);
+		if(dist < 3600.f) {
```
Nick007J: SQR(60.f)

src/audio/AudioManager.cpp
```diff
@@ -8009,13 +9693,11 @@ cAudioManager::SetupJumboRumbleSound(uint8 emittingVol)
 bool
 cAudioManager::SetupJumboTaxiSound(uint8 vol)
 {
-	uint8 emittingVol;
-
 	if(m_sQueueSample.m_fDistance >= 180.f) return 0;
 
-	emittingVol = (vol >> 1) + ((vol >> 1) * m_sQueueSample.m_fDistance * 0.0055556f);
+	uint8 emittingVol = (vol + (vol * m_sQueueSample.m_fDistance / 180)) / 2;
```
Nick007J: technically this is not identical: `vol >> 1` does integer division, and `/2` in the end does float division, might produce minor inaccuracy

src/audio/AudioManager.cpp
```diff
@@ -8200,6 +9882,267 @@ cAudioManager::SetupPedComments(cPedParams *params, uint32 sound)
...
+	CAutomobile *automobile;
+	uint8 color1;
+	int32 index;
+	int32 main_color;
+	int32 sample;
+
+	int32 color_pre_modifier;
+	int32 color_post_modifier;
+
+	constexpr int32 colors[] = {
```
Nick007J: I believe `const static int32` and `constexpr` will be compiled the same way, but aap sure will prefer `static const` :P

src/audio/AudioManager.cpp
```diff
@@ -8200,6 +9882,267 @@ cAudioManager::SetupPedComments(cPedParams *params, uint32 sound)
...
+	    259,  3032, 3032, 259,  3032, 3032, 259,  3032, 3032, 259,  3032, 3032, 259,  3032,
+	    3032, 259,  3032, 3032, 258,  3032, 3032, 258,  3032, 3032, 258,  3032, 3032, 258,
+	    3032, 3032, 258,  3032, 3032};
+
+	if(MusicManager.m_nMusicMode != 2) {
+		automobile = (CAutomobile *)FindPlayerVehicle();
+		if(automobile) {
+			if(60 - policeChannelTimer > 9u) {
+				color1 = automobile->m_currentColour1;
+				if(color1 >= 95u) {
```
Nick007J: I think it can be `sizeof(colours)/sizeof(colours[0])` when you will refactor it as struct

src/control/PathFind.cpp
```diff
@@ -1273,39 +1273,51 @@ CPathFind::DoPathSearch(uint8 type, CVector start, int32 startNodeId, CVector ta
...
+			if(m_pathNodes[startObj->m_nodeIndices[type][i]].group ==
+			   m_pathNodes[targetNode].group)
 				numPathsToTry++;
 		}
-	}else{
+	} else {
 		numPathsToTry = 1;
 		startObj = m_mapObjects[m_pathNodes[startNodeId].objectIndex];
 	}
-	if(numPathsToTry == 0)
```
Nick007J: this goto is one of two use cases that is considered to be acceptable in C. this is C++ of course, but this is C-like code so it makes sense to keep this goto, because it actually becomes worse if you remove goto. aap will not approve

src/core/common.h
```diff
@@ -197,8 +197,8 @@ void re3_assert(const char *expr, const char *filename, unsigned int lineno, con
 #define ARRAY_SIZE(array)                (sizeof(array) / sizeof(array[0]))
 #define BIT(num)                         (1<<(num))
 
-#define max(a, b)  (((a) > (b)) ? (a) : (b)) 
-#define min(a, b)  (((a) < (b)) ? (a) : (b))
+//auto max = [](auto a, auto b) { return ((a) > (b)) ? (a) : (b); };
```
Nick007J: this requires a discussion, because there were no lambdas in C++ that R* used, why would we use them? (other than "it is a better technique now" of course) aap will not approve (x2) also does it even compile without max/min?

src/core/main.cpp
```diff
@@ -339,7 +339,7 @@ void
 RenderEffects(void)
 {
 	CGlass::Render();
-	CWaterCannons::Render();
+	CWaterCannon::Render();
```
Nick007J: why? CWaterCannons::Render() was correct

src/render/WaterCannon.cpp
```diff
@@ -2,5 +2,7 @@
 #include "patcher.h"
 #include "WaterCannon.h"
 
-WRAPPER void CWaterCannons::UpdateOne(uint32 id, CVector *pos, CVector *dir) { EAXJMP(0x522470); }
```
Nick007J: why? it is CWaterCannons

src/render/WaterCannon.h
```diff
@@ -1,8 +1,20 @@
 #pragma once
 
-class CWaterCannons
+class CWaterCannon
```
Nick007J: you confused it. CWaterCannons is a class with static methods without any fields. CWaterCannon is a class with fields, with non-static methods. CWaterCannons::Render calls CWaterCannon::Render. you should have added another class, not replace the old one

src/audio/AudioManager.cpp
```diff
@@ -347,15 +397,63 @@ cAudioManager::AddReflectionsToRequestedQueue()
...
+	bool toProcess[44];
+
+	for(int32 i = 0; i < m_bSampleRequestQueuesStatus[m_bActiveSampleQueue]; i++) {
+		tActiveSample &sample =
+		    m_asSamples[!m_bActiveSampleQueue]
+		               [m_abSampleQueueIndexTable[!m_bActiveSampleQueue][i]];
+		if(!m_asSamples[!m_bActiveSampleQueue]
+		               [m_abSampleQueueIndexTable[!m_bActiveSampleQueue][i]]
+		                   .m_bLoopEnded) {
+			toProcess[i] = 0;
```
Nick007J: =false

src/audio/AudioManager.cpp
```diff
@@ -2954,18 +3148,106 @@ cAudioManager::Initialise()
...
 void
 cAudioManager::InitialisePoliceRadio()
 {
-	EAXJMP(0x57EEC0);
+	policeChannelTimer = 0;
+	policeChannelTimerSeconds = 0;
+	policeChannelCounterSeconds = 0;
+	for(int32 i = 0; i < 10; i++) { crimes[i].type = 0; }
+
+	SampleManager.SetChannelReverbFlag(28, 0);
```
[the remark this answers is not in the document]
ShFil119: yeah, it's police channel

src/audio/AudioManager.cpp
```diff
@@ -529,7 +665,10 @@ cAudioManager::DestroyAllGameCreatedEntities()
 				case AUDIOTYPE_SCRIPTOBJECT:
 					entity =
 					    (cAudioScriptObject *)m_asAudioEntities[i].m_pEntity;
-					if(entity) { delete entity; }
+					if(entity) {
+						delete entity;
+						m_asAudioEntities[i].m_pEntity = nil;
```
Nick007J: this is just `delete entity;` btw, without anything else or any checks

src/core/common.h
ShFil119: MSVC provides function/macro `min`/`max`. 

(It was helper for mingw, so to remove.)

src/audio/AudioManager.cpp
ShFil119: I believe I was getting errors in debug mode about using deleted memory.

src/audio/AudioManager.cpp
ShFil119: `constexpr` is initialized at compile time and has better error checking.

src/audio/AudioManager.cpp
```diff
@@ -347,15 +397,63 @@ cAudioManager::AddReflectionsToRequestedQueue()
...
+				       m_asSamples[m_bActiveSampleQueue]
+				                  [m_abSampleQueueIndexTable[m_bActiveSampleQueue]
+				                                            [j]]
+				                      .m_nEntityIndex &&
+				   sample.m_counter ==
+				       m_asSamples[m_bActiveSampleQueue]
+				                  [m_abSampleQueueIndexTable[m_bActiveSampleQueue]
+				                                            [j]]
+				                      .m_counter) {
+					toProcess[i] = 1;
```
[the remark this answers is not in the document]
ShFil119: done

src/audio/AudioManager.cpp
```diff
@@ -421,6 +522,39 @@ cAudioManager::CheckForAnAudioFileOnCD() const
 	return SampleManager.CheckForAnAudioFileOnCD();
 }
 
+void
+cAudioManager::ClearActiveSamples()
+{
+	for(int32 i = 0; i < m_bActiveSamples; i++) {
+		m_asActiveSamples[i].m_nEntityIndex = -5;
```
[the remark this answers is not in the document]
ShFil119: fixed usages

src/audio/AudioManager.cpp
```diff
@@ -2926,6 +3066,60 @@ cAudioManager::GetPhrase(uint32 *phrase, uint32 *prevPhrase, uint32 sample, uint
...
+	int wheelState;
+	float relativeVelChange;
+	float gasPedalAudio = automobile->m_fGasPedalAudio;
+	float modificator;
+	float velChange;
+	float relativeVel;
+
+	wheelState = automobile->m_aWheelState[wheel];
+	if(wheelState == 1 && gasPedalAudio > 0.4f) {
+		relativeVelChange = (gasPedalAudio + -0.4f) * 1.25f;
```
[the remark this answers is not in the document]
ShFil119: fixed

src/audio/AudioManager.cpp
```diff
@@ -2926,6 +3066,60 @@ cAudioManager::GetPhrase(uint32 *phrase, uint32 *prevPhrase, uint32 sample, uint
...
+{
+	int wheelState;
+	float relativeVelChange;
+	float gasPedalAudio = automobile->m_fGasPedalAudio;
+	float modificator;
+	float velChange;
+	float relativeVel;
+
+	wheelState = automobile->m_aWheelState[wheel];
+	if(wheelState == 1 && gasPedalAudio > 0.4f) {
```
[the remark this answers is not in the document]
ShFil119: fixed usages

src/audio/AudioManager.cpp
```diff
@@ -2954,18 +3148,106 @@ cAudioManager::Initialise()
...
+
+tPoliceRadioZone *ZoneSfx = (tPoliceRadioZone *)0x880240;
+char *SubZo2Label = (char *)0x6E9918;
+char *SubZo3Label = (char *)0x6E9870;
+
 void
 cAudioManager::InitialisePoliceRadioZones()
 {
-	EAXJMP(0x57EAC0);
+	for(int32 i = 0; i < 36; i++) { ZoneSfx[i].m_aName[0] = 0; }
```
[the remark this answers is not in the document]
ShFil119: used loop to fully initialize array

src/audio/AudioManager.cpp
```diff
@@ -2954,18 +3148,106 @@ cAudioManager::Initialise()
...
+
+	strcpy((char *)ZoneSfx, "HOSPI_2");
```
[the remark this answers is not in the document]
ShFil119: Fixed, no macro yet, later I'll think about something better.

src/audio/AudioManager.cpp
```diff
@@ -2954,18 +3148,106 @@ cAudioManager::Initialise()
...
+	ZoneSfx[0].m_nSampleIndex = AUDIO_SAMPLE_POLICE_SCANNER_ZONE_ROCKFORD;
+	strcpy((char *)&ZoneSfx[1], "CONSTRU");
```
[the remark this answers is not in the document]
ShFil119: also fixed

src/audio/AudioManager.cpp
```diff
@@ -3030,18 +3312,297 @@ cAudioManager::PlayLoadedMissionAudio()
 	}
 }
 
-WRAPPER
 void
 cAudioManager::PlayOneShot(int32 index, int16 sound, float vol)
 {
-	EAXJMP(0x57A500);
+	static constexpr uint8 byte_60ABD0[168] = {
```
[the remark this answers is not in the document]
ShFil119: done

src/audio/AudioManager.cpp
ShFil119: reverted

src/audio/AudioManager.cpp
ShFil119: done

src/audio/AudioManager.cpp
ShFil119: fixed

src/audio/AudioManager.cpp
ShFil119: fixed

src/audio/AudioManager.cpp
ShFil119: reverted

src/control/PathFind.cpp
ShFil119: reverted

src/core/common.h
ShFil119: removed

src/audio/AudioManager.cpp
ShFil119: done

src/audio/AudioManager.cpp
```diff
@@ -10245,27 +10245,31 @@ cAudioManager::UsesSirenSwitching(int32 model) const
...
-WRAPPER
 void
-cAudioManager::sub_57C2B0()
+cAudioManager::AdjustSamplesVolume()
 {
-	EAXJMP(0x57C2B0);
+	for(int i = 0; i < m_bSampleRequestQueuesStatus[m_bActiveSampleQueue];
+	    i++) {
+		tActiveSample *pSample =
+		    &m_asSamples[i][(int32)m_abSampleQueueIndexTable[i] + 1];
```
Sergeanur: Why int32 cast?

ShFil119: iirc without it MSVC didn't want to compile file.

PR #255, opened 2019-10-26
script 500-599

Nick007J:


src/core/World.cpp
```diff
@@ -998,6 +998,19 @@ CWorld::RemoveFallenCars(void)
...
 }
 
+void
+CWorld::StopAllLawEnforcersInTheirTracks(void)
+{
+	int poolSize = CPools::GetVehiclePool()->GetSize();
+	for (int poolIndex = poolSize - 1; poolIndex >= 0; poolIndex--) {
+		CVehicle* veh = CPools::GetVehiclePool()->GetSlot(poolIndex);
+		if (veh) {
+			if (veh->bIsLawEnforcer)
```
Sergeanur: `if (veh != nil && veh->bIsLawEnforcer)`

Nick007J: left it consistent with code right above it :P probably might be worth to change them all

PR #267, opened 2019-11-05
cAudioManager::UpdateReflections

ShFil119:


src/audio/AudioManager.cpp
```diff
@@ -9289,11 +9289,75 @@ cAudioManager::UpdateGasPedalAudio(CAutomobile *automobile)
...
+						}
+					}
+				} else {
+					m_avecReflectionsPos[3] = camPos;
+					m_avecReflectionsPos[3].x += 50.f;
+					if(CWorld::ProcessLineOfSight(
+					       camPos, m_avecReflectionsPos[3], colpoint, ent, true,
+					       false, false, true, false, true, true)) {
+						m_afReflectionsDistances[3] =
+						    Distance(camPos, colpoint.point);
```
erorcun: We directly use Magnitude for that. But if you want to add Distance function, you can mimic the one in CVector2D, which takes axises of second vector seperately ( Distance2D(Vector1, x, y) ). If you insist on adding Distance(v1, v2), you should ask it to aap.

src/audio/AudioManager.cpp
```diff
@@ -9289,11 +9289,75 @@ cAudioManager::UpdateGasPedalAudio(CAutomobile *automobile)
...
+				m_afReflectionsDistances[1] = Distance(camPos, colpoint.point);
+			} else {
+				m_afReflectionsDistances[1] = 50.0f;
+			}
+		}
+	} else {
+		m_avecReflectionsPos[0] = camPos;
+		m_avecReflectionsPos[0].y += 50.f;
+		if(CWorld::ProcessLineOfSight(camPos, m_avecReflectionsPos[0], colpoint, ent, 1, 0,
+		                              0, 1, 0, 1, 1)) {
```
erorcun: 0, 1.. -> false, true...

src/audio/AudioManager.cpp
ShFil119: aap said it's ok.

PR #277, opened 2019-11-28
Implement ProcessVehicleSirenOrAlarm

ShFil119:


src/audio/AudioManager.cpp
```diff
@@ -517,11 +517,11 @@ cAudioManager::AgeCrimes()
 }
 
 void
-cAudioManager::CalculateDistance(bool *ptr, float dist)
+cAudioManager::CalculateDistance(bool &condition, float dist)
 {
-	if(*ptr == false) {
+	if(condition == false) {
```
Sergeanur: `if(!condition) {`

PR #281, opened 2019-12-10
CPlayerInfo done

erorcun:


src/core/PlayerInfo.cpp
```diff
@@ -113,8 +118,8 @@ CPlayerInfo::Clear(void)
 	m_nTrafficMultiplier = 0;
 	m_fRoadDensity = 1.0f;
 	m_bInRemoteMode = false;
-	m_bSwitchTaxi = false;
-	m_nSwitchTaxiTime = 0;
+	m_bUnusedTaxiThing = false;
```
Nick007J: There is a script command SWITCH_TAXI_TIMER that uses those fields, don't think it was good either way.

src/core/PlayerInfo.cpp
```diff
@@ -306,6 +311,256 @@ INITSAVEBUF
...
+				}
+			}
+		}
+	}
+}
+
+void
+CPlayerInfo::Process(void)
+{
+	// Unused taxi feature. Gives you a dollar for every second with a passenger. Can be toggled via 0x29A opcode.
```
Nick007J: Never mind you mention it :D

src/core/PlayerInfo.cpp
```diff
@@ -306,6 +311,256 @@ INITSAVEBUF
...
+	if (startTaxiTimer)
+		m_nUnusedTaxiTimer = CTimer::GetTimeInMilliseconds();
+
+	// The effect that makes money counter does while earning/losing money
+	if (m_nVisibleMoney != m_nMoney) {
+		int diff = m_nMoney - m_nVisibleMoney;
+		int diffAbs = Abs(diff);
+		int changeBy;
+
+		if (diffAbs <= 100000) {
```
Nick007J: Almost certain it was
```
if (diffAbs > 100000)
  changeBy = 12345;
else if (diffAbs > 10000)
  changeBy = 1234;
...
```
It looks like that even in assembly, not sure why IDA outputs it like that

erorcun: ok

src/core/PlayerInfo.cpp
erorcun: :)

PR #298, opened 2020-01-27
save

Sergeanur:
All funcs except GenericLoad and GenericSave


erorcun:
Why not add save game struct? At least temporary one until GenericSave/Load got reversed

Sergeanur:
> 
> 
> Why not add save game struct? At least temporary one until GenericSave/Load got reversed

Original code doesn't seem to be using structs, rather stack things up in a file/memory stream manner.

PR #299, opened 2020-01-28
CPedIK finished, bit of CCamera

Sergeanur:


src/peds/PedIK.cpp
```diff
@@ -223,6 +219,151 @@ CPedIK::LookAtPosition(CVector const &pos)
...
+}
+
+bool
+CPedIK::PointGunInDirectionUsingArm(float phi, float theta)
+{
+	bool result = false;
+	RwFrame *frame = m_ped->m_pFrames[PED_UPPERARMR]->frame;
+	RwMatrix *matrix = GetWorldMatrix(RwFrameGetParent(frame), RwMatrixCreate());
+
+	RwV3d upVector = { matrix->right.z, matrix->up.z, matrix->at.z };
```
erorcun: don't we have macros for that? i'm asking because I don't know. `up` and `at` aren't really `up` and `at` as you know.

src/peds/PedIK.cpp
```diff
@@ -223,6 +219,151 @@ CPedIK::LookAtPosition(CVector const &pos)
...
+	else
+		RotateTorso(m_ped->m_pFrames[PED_MID], &m_torsoOrient, false);
+	return result;
+}
+
+bool
+CPedIK::PointGunInDirectionUsingArm(float phi, float theta)
+{
+	bool result = false;
+	RwFrame *frame = m_ped->m_pFrames[PED_UPPERARMR]->frame;
```
erorcun: same

src/peds/PedIK.cpp
```diff
@@ -223,6 +219,151 @@ CPedIK::LookAtPosition(CVector const &pos)
...
+		CGeneral::GetRadianAngleBetweenPoints(position.z, Distance2D(m_ped->GetPosition(), position.x, position.y),
+		m_ped->GetPosition().z,
+		0.0f));
+}
+
+bool
+CPedIK::RestoreLookAt(void)
+{
+	bool result = false;
+	RwMatrix *mat = RwFrameGetMatrix(m_ped->m_pFrames[PED_HEAD]->frame);
```
erorcun: same

src/peds/PedIK.cpp
```diff
@@ -223,6 +219,151 @@ CPedIK::LookAtPosition(CVector const &pos)
...
+
+	return result;
+}
+
+void
+CPedIK::ExtractYawAndPitchWorld(RwMatrixTag *mat, float *yaw, float *pitch)
+{
+	float f = clamp(DotProduct(mat->up, CVector(0.0f, 1.0f, 0.0f)), -1.0f, 1.0f);
+	*yaw = Acos(f);
+	if (mat->up.x > 0.0f) *yaw = -*yaw;
```
erorcun: if we have macros for that you may also consider changing these in Extract... functions to their macro counterparts

src/peds/PedIK.cpp
Sergeanur: Unlikely. Previously reversed funcs were doing the same.

PR #324, opened 2020-02-06
cMusicManager

Sergeanur:


src/audio/MusicManager.cpp
```diff
@@ -147,97 +166,769 @@ cMusicManager::DisplayRadioStationName()
...
+	}
+
+	if (mode2 != m_nMusicMode || mode == MUSICMODE_FRONTEND && mode2 == MUSICMODE_FRONTEND) {
+		switch (mode)
+		{
+		case MUSICMODE_FRONTEND:
+		case MUSICMODE_GAME:
+		case MUSICMODE_CUTSCENE:
+		case MUSICMODE_DISABLED:
+			if (SampleManager.IsStreamPlaying(0)) {
```
ShFil119: enum for stream?

src/audio/MusicManager.cpp
```diff
@@ -147,97 +166,769 @@ cMusicManager::DisplayRadioStationName()
...
+{
+	if (m_bResetTimers) {
+		m_bResetTimers = false;
+		m_nLastTrackServiceTime = m_nResetTime;
+	}
+
+	if (!m_bIsInitialised || m_bDisabled) return;
+
+	if (m_nMusicMode == MUSICMODE_CUTSCENE) {
+		SampleManager.SetStreamedVolumeAndPan(127, 63, 1, 0);
```
ShFil119: 127 could be replaced with maxVolume.

src/audio/MusicManager.cpp
Sergeanur: It's a slot number, no enum for it atm, but it may be a good idea to make one in the future.

src/audio/MusicManager.cpp
```diff
@@ -15,49 +18,65 @@ cMusicManager &MusicManager = *(cMusicManager *)0x8F3964;
...
 cMusicManager::PlayerInCar()
 {
 	if(!FindPlayerVehicle())
 		return false;
-	else {
-		int32 State = CWorld::Players[CWorld::PlayerInFocus].m_pPed->m_nPedState;
 
-		if(State == PED_DRAG_FROM_CAR || State == PED_EXIT_CAR || State == PED_ARRESTED)
-			return false;
+	int32 State = CWorld::Players[CWorld::PlayerInFocus].m_pPed->m_nPedState;
```
erorcun: FindPlayerPed()

PR #328, opened 2020-02-16
CPopulation 1

erorcun:


src/peds/Population.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+void
+CPopulation::LoadPedGroups()
+{
+	int fd;
+	char line[1024];
+	int nextPedGroup = 0;
+
+	// 272?!? what?!?
+	char modelName[272];
```
aap: Probably 256 + 16? Maybe an unused array?

src/peds/Population.cpp
```diff
@@ -0,0 +1,392 @@
...
+WRAPPER void CPopulation::Update(void) { EAXJMP(0x4F39A0); }
+WRAPPER CPed *CPopulation::AddPedInCar(CVehicle *vehicle) { EAXJMP(0x4F5800); }
+WRAPPER bool CPopulation::IsPointInSafeZone(CVector *coors) { EAXJMP(0x4F60C0); }
+
+void
+CPopulation::Initialise()
+{
+	debug("Initialising CPopulation...\n");
+
+	ms_nNumCivMale = 0;
```
Nick007J: From what I noticed, compiler reordered sequential assignments in almost every initializing function, by replacing "1-2-3-4-5-6-7-8" with "1-8-2-7-3-6-4-5" etc. In this case such initialization will make more "sense".

src/peds/Population.cpp
erorcun: ok

src/peds/Population.cpp
erorcun: ok

PR #336, opened 2020-02-24
script utilities 1

Nick007J:


src/control/Script.h
```diff
@@ -388,36 +328,163 @@ class CTheScripts
...
+			retval |= CTheScripts::ScriptSpace[(*pIp)++] << (8 * i);
 		}
 		return retval;
 	}
-	static int8 Read1ByteFromScript(uint32* pIp){
+	static int8 Read1ByteFromScript(uint32* pIp) {
 		int8 retval = 0;
-		for (int i = 0; i < 1; i++){
-			retval |= ScriptSpace[(*pIp)++] << (8 * i);
+		for (int i = 0; i < 1; i++) {
```
ShFil119: not sure if it needs to be loop

ShFil119: not sure if it needs to be loop

src/control/Script.h
```diff
@@ -355,19 +268,43 @@ class CTheScripts
...
+	static void ScriptDebugLine3D(float x1, float y1, float z1, float x2, float y2, float z2, uint32 col, uint32 col2);
+	static void RenderTheScriptDebugLines();
+
+	static void SaveAllScripts(uint8*, uint32*);
+	static void LoadAllScripts(uint8*, uint32);
+
+	static bool IsDebugOn() { return DbgFlag; };
+	static void InvertDebugFlag() { DbgFlag = !DbgFlag; }
+
+	static int32* GetPointerToScriptVariable(int32 offset) { return (int32*)&ScriptSpace[offset]; }
```
ShFil119: asserts?

PR #345, opened 2020-03-11
GenericSave() done

whampson:


src/save/GenericGameStorage.cpp
```diff
@@ -31,9 +48,152 @@ CDate &CompileDateAndTime = *(CDate*)0x72BCB8;
...
+	WriteSaveBuf(buf, (uint32) CClock::ms_nGameClockHours);
+	WriteSaveBuf(buf, (uint32) CClock::ms_nGameClockMinutes);
+	currPad = CPad::GetPad(0);
+	WriteSaveBuf(buf, (uint32) currPad->Mode);
+	WriteSaveBuf(buf, CTimer::m_snTimeInMilliseconds);
+	WriteSaveBuf(buf, CTimer::ms_fTimeScale);
+	WriteSaveBuf(buf, CTimer::ms_fTimeStep);
+	WriteSaveBuf(buf, CTimer::ms_fTimeStepNonClipped);
+	WriteSaveBuf(buf, CTimer::m_FrameCounter);
+	WriteSaveBuf(buf, 1.0f);		// CTimeStep::ms_fTimeStep;
```
Sergeanur: Needs to be replaced with an actual use of CTimeStep

src/save/GenericGameStorage.cpp
```diff
@@ -31,9 +48,152 @@ CDate &CompileDateAndTime = *(CDate*)0x72BCB8;
...
+	WriteSaveBuf(buf, 1.0f);		// CTimeStep::ms_fFramesPerUpdate;
+	WriteSaveBuf(buf, 1.0f);		// CTimeStep::ms_fTimeScale;
```
Sergeanur: Needs to be replaced with an actual use of CTimeStep

Sergeanur:
Faithful to the original code would be use of WriteDataToBufferPointer instead of WriteSaveBuf as it used memcpy to write data.

src/save/GenericGameStorage.cpp
```diff
@@ -31,9 +48,152 @@ CDate &CompileDateAndTime = *(CDate*)0x72BCB8;
...
+
+	// Save simple vars
+INITSAVEBUF
+	lastMissionPassed = TheText.Get(CStats::LastMissionPassedName);
+	if (*lastMissionPassed) {
+		AsciiToUnicode("'...", suffix);
+		TextCopy(saveName, lastMissionPassed);
+		int len = UnicodeStrlen(saveName);
+		saveName[len] = '\0';
+		if (22 < len)
```
Sergeanur: This would be better reversed

whampson: This *is* what the reversed code looks like, as buggy as it may seem (setting the null terminator twice, etc.). Can you elaborate on what you mean?

src/save/GenericGameStorage.cpp
whampson: Will do

src/save/GenericGameStorage.cpp
```diff
@@ -31,9 +48,152 @@ CDate &CompileDateAndTime = *(CDate*)0x72BCB8;
...
+			TextCopy(saveName + 18, suffix);
+		saveName[23] = '\0';
+	}
+	WriteDataToBufferPointer(buf, saveName);
```
whampson: > Faithful to the original code would be use of WriteDataToBufferPointer instead of WriteSaveBuf as it used memcpy to write data.

Where in the code are you referring to? From what I see, the only place where `memcpy` is used is here, where the save name is written to `work_buff`. All the other stuff is written directly to the buffer.

src/save/GenericGameStorage.cpp
whampson: Oh nevermind, I get what you mean, you're saying `if (len > 22)` instead of `if (22 < len)`

src/save/GenericGameStorage.cpp
Sergeanur: Original code is using movsd which is an optimized memcpy. Mobile version still has it unoptimized.

whampson: Makes sense, done.

erorcun:
Merged :) you can check out our Discord server from the link in Readme if you want.

PR #351, opened 2020-03-22
Console

Sergeanur:


src/render/Console.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+	while (m_nLineCount != 0 && CTimer::GetTimeInMilliseconds() - m_aTimer[m_nCurrentLine] > 20000) {
+		m_nLineCount--;
+		m_nCurrentLine = (m_nCurrentLine + 1) % MAX_LINES;
+	}
+
+	for (int16 i = 0; i < m_nLineCount; i++) {
+		int16 line = (i + m_nCurrentLine) % MAX_LINES;
+		CFont::SetColor(CRGBA(0, 0, 0, 200));
+		CFont::PrintString(31.0f, 11.0f + i * 12, Buffers[line]);
```
Nick007J: Maybe a define/const for X/Y offsets? Does it need scaling?

Sergeanur: Added defines. As far as scaling goes I think it's better off without it.

PR #356, opened 2020-03-25
CFire and CFireManager classes nearly done

blingu:


src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+	m_nExtinguishTime = 0;
+	m_nStartTime = 0;
+	field_20 = 1;
+	field_24 = 0;
+	m_fStrength = 0.8f;
+}
+
+CFire::~CFire(void) {}
+
+class CFire_ : public CFire {
```
Nick007J: I think this class isn't needed. CFire is not a virtual class and is never allocated with operator new.

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+
+	fire->m_bIsOngoing = true;
+	fire->m_bIsScriptFire = false;
+	fire->m_vecPos = ped->GetPosition();
+
+	if (entityOnFire && entityOnFire->IsPed() && ped->IsPlayer()) {
+		fire->m_nExtinguishTime = CTimer::m_snTimeInMilliseconds + 3333;
+	} else if (entityOnFire->IsVehicle()) {
+		fire->m_nExtinguishTime = CTimer::m_snTimeInMilliseconds
+			+ 4000 + (signed int)((double)(unsigned __int16)rand() * 0.000030517578 * 1000.0f); //this needs to be simplified
```
Nick007J: CGeneral::GetRandomNumberInRange(4000, 5000)

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+	} else {
+		fire->m_nExtinguishTime = CTimer::m_snTimeInMilliseconds
+			+ 10000 + (signed int)((double)(unsigned __int16)rand() * 0.000030517578 * 1000.0f);
```
Nick007J: same

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+				pCurrentVehicle->InflictDamage((CVehicle *)m_pSource, WEAPONTYPE_FLAMETHROWER, fDamageVehicle);
+			}
+		}
+	}
+	if (!FindPlayerVehicle() && !FindPlayerPed()->m_pFire && !(FindPlayerPed()->bFireProof)
+		&& ((FindPlayerPed()->GetPosition() - m_vecPos).MagnitudeSqr() < 2.0f)) {
+		FindPlayerPed()->DoStuffToGoOnFire();
+		gFireManager.StartFire(FindPlayerPed(), m_pSource, 0.8f, 1);
+	}
+	if (CTimer::m_snTimeInMilliseconds > field_24) { /* set to 0 when a newfire starts, related to time */
```
Nick007J: let's name field_24 maybe? seems like "next time to add flame particles"

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
 CFireManager &gFireManager = *(CFireManager*)0x8F31D0;
 
-WRAPPER void CFire::Extinguish(void) { EAXJMP(0x479D40); }
-WRAPPER void CFireManager::Update(void) { EAXJMP(0x479310); }
-WRAPPER CFire* CFireManager::FindFurthestFire_NeverMindFireMen(CVector coors, float, float) { EAXJMP(0x479430); }
+CFire::CFire(void)
+{
+	m_bIsOngoing = 0;
+	m_bIsScriptFire = 0;
+	m_bPropagationFlag = 1;
```
erorcun: let's make bool's only true/false

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+	float fDamageVehicle;
+	CVehicle *pCurrentVehicle;
+	char nRandNumber;
+	float fGreen;
+	float fRed;
+	CVector lightpos;
+	CVector firePos;
+	CVector vecProduct;
+
+	CEntity *pCurrentEntity = (CPed *)m_pEntity;
```
erorcun: you already have m_pEntity and don't need creating new variable. please don't rely on IDA output in these cases, it doesn't represent the original code, so local variables are mostly up to you.

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+		rand(); rand(); rand(); /* unsure why these three rands are called */
+
+		CParticle::AddParticle(PARTICLE_CARFLAME_SMOKE, firePos,
+			CVector(0.0f, 0.0f, 0.0f), 0, 0.0, 0, 0, 0, 0);
+	}
+	if (CTimer::m_snTimeInMilliseconds < m_nExtinguishTime || m_bIsScriptFire) {
+		if (CTimer::m_snTimeInMilliseconds > m_nStartTime)
+			m_nStartTime = CTimer::m_snTimeInMilliseconds + 400;
+
+		nRandNumber = rand() & 127;
```
erorcun: CGeneral::GetRandomNumber()

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+	if (m_bIsOngoing) {
+		if (!m_bIsScriptFire)
+			gFireManager.m_nTotalFires--;
+
+		m_nExtinguishTime = 0;
+		m_bIsOngoing = false;
+		pCurrentEntity = (CPed *)m_pEntity;
+
+		if (pCurrentEntity) {
+			if (pCurrentEntity->IsObject()) {
```
erorcun: this is IsPed()

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+				pCurrentEntity->RestorePreviousState();
+				pCurrentEntity->m_pFire = 0;
+			} else if (pCurrentEntity->IsPed()) {
```
erorcun: this is IsVehicle(), and the variable below is m_pCarFire

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+{
+	CPed *ped = (CPed *)entityOnFire;
+	CVehicle *veh = (CVehicle *)entityOnFire;
+	CFire *fire = GetNextFreeFire();
+
+	if (!fire)
+		return;
+	if (entityOnFire->IsPed() && (ped->m_pFire || !ped->IsPedInControl())) {
+		return;
+	} else if (entityOnFire->IsVehicle() && veh->m_pCarFire || veh->IsCar() && ((CAutomobile *)veh)->Damage.GetEngineStatus() >= 225) {
```
erorcun: parenthesis are placed wrong

src/core/Fire.cpp
```diff
@@ -1,19 +1,299 @@
...
+		pCurrentFire->m_fStrength = size;
+	}
+}
+
+void
+CFireManager::StartFire(CEntity *entityOnFire, CEntity *fleeFrom, float strength, bool propagation)
+{
+	CPed *ped = (CPed *)entityOnFire;
+	CVehicle *veh = (CVehicle *)entityOnFire;
+	CFire *fire = GetNextFreeFire();
```
erorcun: game runs GetNextFreeFire()  after the entity type checks below, let's do the same

src/core/Fire.cpp
blingu: so remove CFire_ and InjectHook to ctor/dtor?

src/core/Fire.cpp
blingu: isn't this fireDuration? I'm unsure about this so I leave it to you.

src/core/Fire.cpp
blingu: go more into detail please

erorcun: move GetNextFreeFire() below return 0's

src/core/Fire.cpp
blingu: which ones do you mean?

erorcun: actually it's missing. entityOnFire->IsVehicle() should be above of other checks.

src/core/Fire.cpp
Nick007J: yeah, you can just remove it

src/core/Fire.cpp
Nick007J: well it isn't fire duration because it's updated by arbitrary number which is then being checked against

src/core/Fire.cpp
```diff
@@ -135,8 +128,7 @@ CFire::ProcessFire(void)
 		if (CTimer::m_snTimeInMilliseconds > m_nStartTime)
 			m_nStartTime = CTimer::m_snTimeInMilliseconds + 400;
 
-		nRandNumber = rand() & 127;
-
+		nRandNumber = CGeneral::GetRandomNumber();
```
erorcun: you still need & 127 tho

src/core/Fire.cpp
```diff
@@ -166,22 +158,19 @@ CFire::ReportThisFire(void)
...
-			if (pCurrentEntity->IsObject()) {
-				pCurrentEntity->RestorePreviousState();
-				pCurrentEntity->m_pFire = 0;
-			} else if (pCurrentEntity->IsPed()) {
-				pCurrentEntity->m_vecOffsetSeek.x = 0.0;
+
+		if (m_pEntity) {
+			if (m_pEntity->IsPed()) {
+				((CPed *)m_pEntity)->RestorePreviousState();
+				((CPed *)m_pEntity)->m_pFire = 0;
```
erorcun: pls also do that nil

src/core/Fire.cpp
```diff
@@ -49,16 +357,100 @@ CFireManager::ExtinguishPoint(CVector point, float range)
...
+			veh->m_pCarFire = fire;
+		}
+	}
+	return fire - m_aFires;
+}
+
+bool
+CFireManager::IsScriptFireExtinguish(int16 index)
+{
+	return (!m_aFires[index].m_bIsOngoing) ? true : false;
```
ShFil119: isn't it over engineered?
`return !m_aFires[index].m_bIsOngoing;`

blingu: i can make it less complicated for you (:

src/core/Fire.cpp
```diff
@@ -49,16 +357,100 @@ CFireManager::ExtinguishPoint(CVector point, float range)
...
+			veh->m_pCarFire = fire;
+		}
+	}
+	return fire - m_aFires;
+}
+
+bool
+CFireManager::IsScriptFireExtinguish(int16 index)
+{
+	return m_aFires[index].m_bIsOngoing ? false : true;
```
Nick007J: return !m_aFires[index].m_bIsOngoing

ShFil119: yeah

blingu: done

src/core/Fire.cpp
```diff
@@ -1,19 +1,289 @@
...
+
+		nRandNumber = CGeneral::GetRandomNumber() & 127;
+		lightpos.x = m_vecPos.x;
+		lightpos.y = m_vecPos.y;
+		lightpos.z = m_vecPos.z + 5.0f;
+
+		if (!m_pEntity) {
+			CShadows::StoreStaticShadow((uint32)this, SHADOWTYPE_ADDITIVE, gpShadowExplosionTex, &lightpos,
+				7.0, 0.0, 0.0, -7.0, 0, nRandNumber / 2, nRandNumber / 2,
+				0, 10.0, 1.0, 40.0, 0, 0.0);
```
ShFil119: floats instead of doubles

src/core/Fire.cpp
```diff
@@ -38,6 +308,44 @@ CFire* CFireManager::FindNearestFire(CVector vecPos, float* pDistance)
 	return nil;
 }
 
+CFire *
+CFireManager::FindFurthestFire_NeverMindFireMen(CVector coords, float minRange, float maxRange)
+{
+	int furthestFire = -1;
+	float lastFireDist = 0.0;
```
ShFil119: also

PR #368, opened 2020-03-29
rotating FollowPed cam (disabled by default for now)

aap:


src/core/Cam.cpp
```diff
@@ -4369,7 +4375,209 @@ CCam::Process_FollowPed_WithBinding(const CVector &CameraTarget, float TargetOri
...
+		LookLeftRight = -CPad::GetPad(0)->LookAroundLeftRight();
+		LookUpDown = CPad::GetPad(0)->LookAroundUpDown();
+	}
+	float AlphaOffset, BetaOffset;
+	if(UseMouse){
+		BetaOffset = LookLeftRight * TheCamera.m_fMouseAccelHorzntl * FOV/80.0f;
+		AlphaOffset = LookUpDown * TheCamera.m_fMouseAccelVertical * FOV/80.0f;
+	}else{
+		BetaOffset = LookLeftRight * fStickSens * (0.5f/10.0f) * FOV/80.0f * CTimer::GetTimeStep();
+		AlphaOffset = LookUpDown * fStickSens * (0.3f/10.0f) * FOV/80.0f * CTimer::GetTimeStep();
```
ShFil119: `(0.5f/10.0f`) vs `/ 20.f`
`(0.3f/10.0f)` vs `* 3.f / 100.f `

aap: I used the same calculation from the other function but changed 7 to 10, so this is ok.

PR #375, opened 2020-03-31
AudioScriptObject finished

Sergeanur:


src/audio/AudioScriptObject.h
```diff
@@ -2,130 +2,132 @@
...
-	_SCRSOUND_GLASS_CRACK  = 116,
-	_SCRSOUND_GLASS_SHARD  = 117,
-	_SCRSOUND_WOODEN_BOX_SMASH  = 118,
-	_SCRSOUND_CARDBOARD_BOX_SMASH  = 119,
-	_SCRSOUND_COL_CAR   = 120,
-	_SCRSOUND_TYRE_BUMP  = 121,
-	_SCRSOUND_BULLET_SHELL_HIT_GROUND_1  = 122,
-	_SCRSOUND_BULLET_SHELL_HIT_GROUND_2  = 123,
+	SCRSOUND_TEST_1,
+	_SCRSOUND_UNK_1,
```
ShFil119: can we avoid underscores in names?

src/audio/AudioScriptObject.cpp
```diff
@@ -19,22 +17,66 @@ cAudioScriptObject::operator new(size_t sz)
...
 	return CPools::GetAudioScriptObjectPool()->New();
 }
+
 void *
 cAudioScriptObject::operator new(size_t sz, int handle)
 {
 	return CPools::GetAudioScriptObjectPool()->New(handle);
 }
 void
+
```
ShFil119: I think it should be above.

PR #377, opened 2020-04-01
Glass

Fire-Head:


src/render/Glass.cpp
```diff
@@ -1,21 +1,721 @@
...
 
-WRAPPER void
-CGlass::WindowRespondsToCollision(CEntity *ent, float amount, CVector speed, CVector point, bool foo)
+uint32 CGlass::NumGlassEntities;
+CEntity *CGlass::apEntitiesToBeRendered[NUM_GLASSENTITIES];
+CFallingGlassPane CGlass::aGlassPanes[NUM_GLASSPANES];
+
+
+CVector2D CentersWithTriangle[NUM_GLASSTRIANGLES];
+CVector2D CoorsWithTriangle[NUM_GLASSTRIANGLES][3] =
```
ShFil119, line 24: To show it's hardcoded data, it could be marked as `const`.

src/render/Glass.cpp
```diff
@@ -1,21 +1,721 @@
...
+
+			pos = CVector(GetPosition().x, GetPosition().y, m_fGroundZ);
+
+			PlayOneShotScriptObject(_SCRSOUND_GLASS_SHARD, pos);
+
+			RwRGBA color = { 255, 255, 255, 255 };
+
+			static int32 nFrameGen = 0;
+
+			for ( int32 i = 0; i < 4; i++ )
```
ShFil119, line 110: in config.h max particles of glass?

src/render/Glass.cpp
```diff
@@ -1,21 +1,721 @@
...
+	if ( upSteps < 1.0f ) upSteps = 1.0f;
+
+	float rightSteps = rightLen + 0.75f;
+	if ( rightSteps < 1.0f ) rightSteps = 1.0f;
+
+	uint32 ysteps = (uint32)upSteps;
+	if ( ysteps > 3 ) ysteps = 3;
+
+	uint32 xsteps = (uint32)rightSteps;
+	if ( xsteps > 3 ) xsteps = 3;
```
ShFil119, line 310: min/max?

src/render/Glass.cpp
```diff
@@ -1,21 +1,721 @@
...
+CFallingGlassPane *
+CGlass::FindFreePane(void)
+{
+	for ( int32 i = 0; i < NUM_GLASSPANES; i++ )
+	{
+		if ( !aGlassPanes[i].m_bActive )
+			return &aGlassPanes[i];
+	}
+
+	return NULL;
```
ShFil119, line 290: nil

erorcun:
lgtm

src/render/Glass.cpp
Fire-Head: it's not looking like min/max, don't think r* used it here

src/render/Glass.cpp
Fire-Head: nah

PR #379, opened 2020-04-04
Frontend (last-3)

erorcun:


src/core/Radar.cpp
```diff
@@ -1105,18 +1154,33 @@ void CRadar::TransformRadarPointToRealWorldSpace(CVector2D &out, const CVector2D
...
 #endif
-	out.y = (1.0f - in.y)*0.5f*SCREEN_SCALE_Y(RADAR_HEIGHT) + SCREEN_SCALE_FROM_BOTTOM(RADAR_BOTTOM + RADAR_HEIGHT);
+		out.y = (1.0f - in.y) * 0.5f * SCREEN_SCALE_Y(RADAR_HEIGHT) + SCREEN_SCALE_FROM_BOTTOM(RADAR_BOTTOM + RADAR_HEIGHT);
+	}
 }
 
 void CRadar::TransformRealWorldPointToRadarSpace(CVector2D &out, const CVector2D &in)
 {
 	float s, c;
+#if 1
```
Sergeanur, line 1177: #ifdef MENU_MAP?

src/core/Radar.cpp
```diff
@@ -1075,6 +1118,11 @@ void CRadar::TransformRealWorldToTexCoordSpace(CVector2D &out, const CVector2D &
 void CRadar::TransformRadarPointToRealWorldSpace(CVector2D &out, const CVector2D &in)
 {
 	float s, c;
+#if 1
```
Sergeanur, line 1121: Same

src/core/PlayerSkin.cpp
```diff
@@ -142,10 +142,6 @@ CPlayerSkin::RenderFrontendSkinEdit(void)
 	const RwV3d axis2 = { 0.0f, 0.0f, 1.0f };
 	static uint32 LastFlash = 0;
 
-#ifdef ASPECT_RATIO_SCALE
-	pos.x = 1.35f * (SCREEN_ASPECT_RATIO / DEFAULT_ASPECT_RATIO);
-#endif
```
Sergeanur, line 147: Why?

src/core/Radar.cpp
erorcun: this is always cached in other games independent from menu map, so i thought why not

src/core/Radar.cpp
```diff
@@ -685,6 +700,19 @@ void CRadar::DrawRadarSection(int32 x, int32 y)
 void CRadar::DrawRadarSprite(uint16 sprite, float x, float y, uint8 alpha)
 {
 	RadarSprites[sprite]->Draw(CRect(x - SCREEN_SCALE_X(8.0f), y - SCREEN_SCALE_Y(8.0f), x + SCREEN_SCALE_X(8.0f), y + SCREEN_SCALE_Y(8.0f)), CRGBA(255, 255, 255, alpha));
+#ifdef MENU_MAP
+	if (bMenuMapActive) {
+		bool alreadyThere;
```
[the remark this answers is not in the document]
erorcun: this code hadn't been run even once at that point, I changed many things in local and will push it when I will enable that define.

src/core/PlayerSkin.cpp
erorcun: without that code but with our changes to rendering, player was already being rendered at the middle-left(this is how it supposed it be, since our menu widecreen fix wraps all the content to middle), this code was there to move player to left-most, since player setup screen was stretching elements to whole screen before this commit.

PR #384, opened 2020-04-06
Menu map

erorcun:


src/core/Camera.h
```diff
@@ -23,6 +23,16 @@ enum
 #define CAM_ZOOM_TOPDOWN (4.0f)
 #define CAM_ZOOM_CINEMATIC (5.0f)
 
+#ifdef FREE_CAM // LCS values
```
aap, line 26: This needs to check bFreeCam somewhere as well.

erorcun: ok

src/core/Camera.cpp
```diff
@@ -2956,11 +2968,23 @@ CCamera::SetZoomValueFollowPedScript(int16 dist)
...
+		case 2: m_fCarZoomValueScript = ((CVehicle*)Cams[ActiveCam].CamTargetEntity)->IsBoat() ? FREE_CAM_BOAT_ZOOM_3 : FREE_CAM_CAR_ZOOM_3; break;
+		default: m_fCarZoomValueScript = m_fCarZoomValueScript; break;
+		}
+	} else
+#endif
+	{
+		switch (dist) {
+		case 0: m_fCarZoomValueScript = 0.05f; break;
+		case 1: m_fCarZoomValueScript = 1.9f; break;
+		case 2: m_fCarZoomValueScript = 3.9f; break;
```
ShFil119, line 2985:
```
#define CAR_CAM_ZOOM_1 0.05f
#define CAR_CAM_ZOOM_2 1.9f
#define CAR_CAM_ZOOM_3 3.9f
```
from header

ShFil119: btw here and above could be used enum

erorcun: enum with floats? :D
btw CAR_CAM_ZOOM_1 things were leftover, thanks for reminding

erorcun: ok, but if you mean changing CAM_ZOOM_x into enum you can ask that to aap, that's not my code

ShFil119: I thought about switch cases

PR #387, opened 2020-04-08
implemented CSkidmarks

aap:


src/render/Skidmarks.cpp
```diff
@@ -1,12 +1,247 @@
...
 #include "Skidmarks.h"
 
-WRAPPER void CSkidmarks::Clear(void) { EAXJMP(0x518130); }
-WRAPPER void CSkidmarks::Update() { EAXJMP(0x518200); }
+CSkidmark CSkidmarks::aSkidmarks[NUMSKIDMARKS];
 
-WRAPPER void CSkidmarks::Render(void) { EAXJMP(0x5182E0); }
-WRAPPER void CSkidmarks::RegisterOne(uint32 id, CVector pos, float fwdx, float fwdY, bool *isMuddy, bool *isBloddy) { EAXJMP(0x5185C0); }
+RwImVertexIndex SkidmarkIndexList[16 * 6];
+RwIm3DVertex SkidmarkVertices[16 * 2];
```
ShFil119, line 12: aren't they two dimensions arrays?

aap: the RW functions take pointers...so i'd say no. but semantically yes.

ShFil119: btw names for sizes (16, 6 and 2) would be nice, these values are used for loops.

aap: eh, ok. we have SKIDMARK_LENGTH now.

PR #389, opened 2020-04-08
Weather

Nick007J:


src/render/Weather.cpp
```diff
@@ -53,6 +259,258 @@ void CWeather::ForceWeatherNow(int16 weather)
...
+			}
+			else
+				Streaks[i].position += (TheCamera.GetPosition() - TheCamera.m_RealPreviousCameraPosition) * 20.0f;
+			Streaks[i].position.x += ((CGeneral::GetRandomNumber() & 255) - 128) * 0.08f;
+			Streaks[i].position.y += ((CGeneral::GetRandomNumber() & 255) - 128) * 0.08f;
+			Streaks[i].timer = CTimer::GetTimeInMilliseconds();
+		}
+	}
+	if (TempBufferIndicesStored){
+		RwRenderStateSet(rwRENDERSTATEZWRITEENABLE, 0);
```
aap, line 490: use enums here. FALSE/TRUE for booleans and rw enums for the others.

PR #390, opened 2020-04-08
Finished CExplosion

Sergeanur:


src/render/Rubbish.cpp
```diff
@@ -230,15 +230,15 @@ CRubbish::Update(void)
 		spawnDist = (CGeneral::GetRandomNumber()&0xFF)/256.0f + RUBBISH_MAX_DIST;
 		uint8 r = CGeneral::GetRandomNumber();
 		if(r&1)
-			spawnAngle = (CGeneral::GetRandomNumber()&0xFF)/256.0f * 6.28f;
+			spawnAngle = (CGeneral::GetRandomNumber()&0xFF)/256.0f * TWOPI;
```
ShFil119, line 233: maybe adding `#if ALWAYS_ACCURATE_PI ` ?

aap: we should keep it as 6.28 for now and if we decide to get rid of 6.28 we'll do it everywhere in one PR and not piecemeal.

PR #394, opened 2020-04-09
CStats finished

Sergeanur:


ShFil119:
Needs rebasing.

src/core/Stats.cpp
```diff
@@ -200,6 +200,229 @@ void CStats::SetTotalNumberMissions(int32 total)
...
+
+	rating = FiresExtinguished + 10 * HighestLevelAmbulanceMission + CriminalsCaught + LivesSavedWithAmbulance
+		+ 30 * HelisDestroyed + TotalLegitimateKills - 3 * TimesArrested - 3 * TimesDied
+		+ CWorld::Players[CWorld::PlayerInFocus].m_nMoney / 5000;
+	if (rating <= 0) rating = 0;
+
+	if (InstantHitsFiredByPlayer > 100)
+		rating += (double)CStats::InstantHitsHitByPlayer / (double)CStats::InstantHitsFiredByPlayer * 500.0f;
+	if (TotalProgressInGame)
+		rating += (double)CStats::ProgressMade / (double)CStats::TotalProgressInGame * 1000.0f;
```
ShFil119, line 236: mobile is using floats here. not sure if we need doubles

erorcun:
lgtm

src/core/Stats.cpp
aap: I assume this is a case of missing f-suffix on the constants. everything gets promoted to double in that case. We should just make it float.

PR #414, opened 2020-04-13
Finish audio

ShFil119:


src/audio/AudioManager.cpp
```diff
@@ -40,19 +40,18 @@
 #include "sampman.h"
 
 cAudioManager &AudioManager = *(cAudioManager *)0x880FC0;
```
erorcun, line 42: why not remove this wrapper :)

ShFil119: I thought that maybe there's unreversed function (yet) which uses audio.

erorcun: I just looked to xref's to AudioManager, but ofc you know better in that regard

erorcun:
lgtm

PR #448, opened 2020-04-15
CBulletInfo

Nick007J:


src/weapons/BulletInfo.cpp
```diff
@@ -2,8 +2,268 @@
...
+	if (minY < y2 || maxY > y1) {
+		if (minY < y1)
+			minP = min(minP, (y1 - minY) / (maxY - minY));
+		if (maxY > y2)
+			maxP = max(maxP, (maxY - y2) / (maxY - minY));
+	}
+#ifdef FIX_BUGS
+	else
+		return false;
+#endif
```
Sergeanur, line 256: This is in #else already

Nick007J: I know, I left it in case if someone wants to fix it closely to original, if it even has any chance of being done. It was certainly a bug on its own - but not the only one.

erorcun:
lgtm

PR #454, opened 2020-04-16
Controller menu

Sergeanur:
Adds controller configuration menu with ability to change setup and set vibration on and off. 
изображение


IlDucci:
I'm going to continue the non-PS2 controller images discussion from https://github.com/GTAmodding/re3/issues/845 here, just so both issues are on topic.

I've done the first pictures, for now for the XBOX 360 controller (this is the end file, I'm willing to share the PSD file. It's vectorized, so it can be resized into HD or higher)
fe_controller_360

There's a couple of issues that I'd like to comment. The biggest one is purely physical: buttons are not in the same place as the PS2 controller, so the arrow positioning won't work just like that. The way I see it, there's two ways to solve this: the artist-based way, and the programming-based way.

The artist-based way would consist in adding new positions for the text via code, following the new positions of the arrows. It's visually cleaner.
fe_controller_360_CUSTOMPERCONTROLLER

And the programmer-based way would respect the original starting position of the arrows, but would look a bit more confusing (specially when the left stick and the directional pad are reversed, just like in this case).
fe_controller_360_PRESERVINGPS2POSITIONS

What do you think about all of this?

P.D. Just realized there seems to be an action missing on presets 3 and 4, where the arrow that points towards both L2 and R2 should have the "Look back" action.

PR #462, opened 2020-04-17
debug menu and weather edit

Juarez12:


erorcun:
lgtm

Sergeanur:
CTimeCycle::Save needs to be ifdef'ed

aap:
Do we actually need that tga reader? i don't think so...

aap:
I'll close this now because I implemented the debugmenu myself using RtCharset. For weather edit you can make a separate PR.

PR #475, opened 2020-04-19
Struct cleanup, various fixes, enable PS2 rand

erorcun:


src/audio/sampman.h
```diff
@@ -45,7 +45,7 @@ class cSampleManager
 	uint8   m_nEffectsFadeVolume;
 	uint8   m_nMusicFadeVolume;
 	uint8   m_nMonoMode;
-	char _pad0[1];
+	char	unk;
```
Fire-Head, line 48: not an unknown, it's struct alignment(padding) better to remove this one or leave as it was before

erorcun: I removed it and checked offset of other elements, they changed. so it should not be padding.

PR #477, opened 2020-04-19
Try to make code working on more platforms

ShFil119:
What's done:
- removed some behavior which is extension made by ms (like predefined enums)
- fixed case sensitivity issues (msvc doesn't care)
- removed some goto 
- renamed min/max to Min/Max

What's changed:
- it's possible now compile re3 with clang-cl on windows (but you need to use default linker)
- using gcc is not possible, because there's still no alternative for many lines of windows like code
- mingw doesn't like fakerw (need rechecking)


Sergeanur:
Does eLevelName needs a separate file? Game.h is included into Streaming.h anyway.

erorcun:
lgtm

ShFil119:
@Sergeanur, reverted.

PR #492, opened 2020-04-26
Initial GLFW support

erorcun:
There are 2 important TODO,

1- Fix "libcmt.lib" conflict warning
2- Fix APIENTRY warning

win.h should not be included anywhere manually now.

### DIFF
--- a/src/core/ControllerConfig.h
+++ b/src/core/ControllerConfig.h
@@ -96,6 +96,16 @@ class CControllerState;
 
 #define ACTIONNAME_LENGTH 40
 
+#ifdef RW_GL3
+struct GlfwJoyState {
+	int8 id;
+	bool isGamepad;
+	uint8 numButtons;
+	uint8* buttons;
+	bool mappedButtons[17];
+};
+#endif
+
 class CControllerConfigManager
 {
 public:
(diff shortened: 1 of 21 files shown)


src/core/ControllerConfig.h
```diff
@@ -115,8 +125,9 @@ class CControllerConfigManager
 #ifdef __DINPUT_INCLUDED__
 	DIJOYSTATE2           m_OldState;
 	DIJOYSTATE2           m_NewState;
-#else
-	uint32 ___padd[0x110 / 4 * 2];
```
Sergeanur, line 119: This was here to avoid compilation error and preserve size in cpp's where windows stuff is not included. Is that fixed now?

src/core/ControllerConfig.h
```diff
@@ -193,6 +204,6 @@ class CControllerConfigManager
 	void  ResetSettingOrder                   (e_ControllerAction action);
 };
 
-VALIDATE_SIZE(CControllerConfigManager, 0x143C);
+//VALIDATE_SIZE(CControllerConfigManager, 0x143C);
```
Sergeanur, line 207: Safe to leave within ifdef for now.

src/core/config.h
```diff
@@ -200,7 +200,7 @@ enum Config {
 //#define PS2_ALTERNATIVE_CARSPLASH // unused on PS2
 
 // Pad
-#define XINPUT
+// #define XINPUT
```
Sergeanur, line 203: Ifdef maybe?

PR #495, opened 2020-04-26
Japanese support

Sergeanur:


src/core/Game.cpp
```diff
@@ -97,8 +97,8 @@ bool CGame::noProstitutes;
 bool CGame::playingIntro;
 char CGame::aDatFile[32];
 #ifdef MORE_LANGUAGES
-bool CGame::polishGame = false;
```
erorcun, line 100: no polish? :c

Noskillx:
Imo bring Polish back and then can be merged.

src/core/Game.cpp
Sergeanur: Unused variable

erorcun: oh ok, lgtm then

PR #507, opened 2020-05-03
implemented most of vice city path system, conditionals for train, bridge, cullzones

aap:
define MIAMI to use this
roadblocks are stubbed out for the moment but they don't work without proper path data anyway.


erorcun:
lgtm

src/control/PathFind.cpp
```diff
@@ -765,14 +1163,19 @@ CPathFind::PreparePathDataForType(uint8 type, CTempNode *tempnodes, CPathInfoFor
 		done = 0;
 		// Set number of lanes for all nodes somehow
 		// very strange code
+#ifndef MIAMI
 		for(k = 0; !done && k < 10; k++){
+#else
+		for(k = 0; !done && k < 12; k++){
+#endif
```
ShFil119, line 1170: ARRAY_SIZE would be nicer

aap: there is no array.

src/control/PathFind.cpp
```diff
@@ -1594,20 +2154,27 @@ static CPathNode *pNodeList[32];
...
 	if(type == PATH_CAR)
 		DoPathSearch(type, start, -1, target, pNodeList, &DummyResult, 32, nil, &dist, 999999.88f, -1);
 	else
 		DoPathSearch(type, start, -1, target, nil, &DummyResult2, 0, nil, &dist, 50.0f, -1);
 	if(type == PATH_CAR)
+#ifndef MIAMI
 		return dist < 160.0f;
+#else
+		return dist < 150.0f;
+#endif
```
ShFil119, line 2172: for this type of could I think: define wrapped with ifndef macro would be best.

aap: agreed, but MIAMI is meant to be temporary anyway, so i think we can leave it like this until we split the code.

src/control/PathFind.cpp
ShFil119: so define?

PR #516, opened 2020-05-05
miami (script + other) stuff

Nick007J:


Sergeanur:
We need IsStatic() to be in master too.

PR #554, opened 2020-05-16
Gcc warnings ptr to int

ShFil119:


src/core/Zones.h
```diff
@@ -106,7 +106,7 @@ class CTheZones
 	static int16 FindAudioZone(CVector *pos);
 	static eLevelName FindZoneForPoint(const CVector &pos);
 	static CZone *GetPointerForZoneIndex(int32 i) { return i == -1 ? nil : &ZoneArray[i]; }
-	static int32 GetIndexForZonePointer(CZone *zone) { return zone == nil ? -1 : zone - ZoneArray; }
+	static uintptr GetIndexForZonePointer(CZone *zone) { return zone == nil ? -1 : zone - ZoneArray; }
```
blingu, line 109: I think an unsigned return type doesn't belong here, as it's possible to return -1

rwsdk/include/d3d8/rwplcore.h
```diff
@@ -1567,7 +1567,7 @@ struct RwSurfaceProperties
 #define rwPIOVER2 (rwPI / (RwReal)(2.0f))
 #endif
 #define RWRGBALONG(r,g,b,a)                                             \
-    ((RwUInt32) (((a) << 24) | ((r) << 16) | ((g) << 8) | (b)))
```
aap, line 1570: no

src/core/Zones.h
aap: agreed. int32 is correct here. technically ssize_t would be the correct type but that's not portable and we don't have 2 billion zones anyway

src/core/templates.h
```diff
@@ -127,9 +127,9 @@ class CPool
 		int i = GetJustIndex(entry);
 		return m_flags[i].u + (i<<8);
 	}
-	int GetJustIndex(T *entry){
+	uintptr GetJustIndex(T *entry){
 		// TODO: the cast is unsafe
-		return (int)((U*)entry - m_entries);
+		return (uintptr)((U*)entry - m_entries);
```
aap, line 132: this is size_t

src/rw/VisibilityPlugins.cpp
```diff
@@ -211,7 +211,7 @@ CVisibilityPlugins::RenderObjNormalAtomic(RpAtomic *atomic)
 }
 
 RpAtomic*
-CVisibilityPlugins::RenderAlphaAtomic(RpAtomic *atomic, int alpha)
+CVisibilityPlugins::RenderAlphaAtomic(RpAtomic *atomic, uintptr alpha)
```
aap, line 214: what? why?

src/rw/VisibilityPlugins.cpp
```diff
@@ -231,7 +231,7 @@ CVisibilityPlugins::RenderFadingAtomic(RpAtomic *atomic, float camdist)
 {
 	RpAtomic *lodatm;
 	float fadefactor;
-	uint32 alpha;
+	uintptr alpha;
```
aap, line 234: again, why?

src/rw/VisibilityPlugins.cpp
```diff
@@ -802,7 +802,7 @@ CVisibilityPlugins::SetFrameHierarchyId(RwFrame *frame, int32 id)
 	FRAMEEXT(frame)->id = id;
 }
 
-int32
+uintptr
 CVisibilityPlugins::GetFrameHierarchyId(RwFrame *frame)
```
aap, line 806: no. this is an int32

src/vehicles/Automobile.cpp
```diff
@@ -4542,7 +4542,7 @@ SetVehicleAtomicVisibilityCB(RwObject *object, void *data)
 }
 
 void
-CAutomobile::SetComponentVisibility(RwFrame *frame, uint32 flags)
+CAutomobile::SetComponentVisibility(RwFrame *frame, uintptr flags)
```
aap, line 4545: also wrong. it's an int

aap:
this is all wrong

src/core/Zones.h
erorcun: intptr_t is portable - or re3 has definitions of it

PR #574, opened 2020-05-21
many additions

Xinerki:
here we go


src/core/Frontend.cpp
```diff
@@ -3115,6 +3116,9 @@ CMenuManager::LoadSettings()
 			CFileMgr::Read(fileHandle, m_PrefsSkinFile, 256);
 			CFileMgr::Read(fileHandle, (char*)&m_ControlMethod, 1);
 			CFileMgr::Read(fileHandle, (char*)&m_PrefsLanguage, 1);
+#ifdef FREE_CAM
+			CFileMgr::Read(fileHandle, (char*)&TheCamera.bFreeCam, 1);
```
erorcun, line 3120: pls keep it in junk data. you can see an example above (prefsWidth/Height)

erorcun: turns out our CFileMgr::Read silently handles the missing field and initializes the variablr to 0.

PR #575, opened 2020-05-21
Small improvements to Pickup and main menu texts

Bigbossbro08:
I changed names as much as I was able to reproduce. Some might be left here and there. Please tell me and I will try my best to fix it.
I also did improvements to pickups. There should no corona around pickups. Atleast for healths. I couldn't do for weapons. Because I couldn't find the exact code for them.


erorcun:
that pickup code isn't VC code

Bigbossbro08:
> 
> 
> that pickup code isn't VC code

I looked into VC code. It was one big if else tree. It looked overall code is same except GTA VC got no glow. Filtered by alpha value as in the code. What I should do then?

erorcun:
VC doesn't completely remove glow in health/armour etc. pickups, just gives them inner glow.
it also sets colorId for them, even though they're unused.
I did it like VC in https://github.com/GTAmodding/re3/pull/634 ,(but outer glow of other pickups waiting CPickup to be done by someone else) so I'm closing this PR.

PR #587, opened 2020-05-24
OPUS support for OpenAL

Sergeanur:
Audio converted to opus: http://www.mediafire.com/file/baeqqfnemur1q8b/opus.7z/file


.gitmodules
```diff
@@ -1,3 +1,15 @@
...
+	url = https://github.com/xiph/opus.git
+	branch = master
+[submodule "opusfile"]
+	path = opusfile
+	url = https://github.com/xiph/opusfile.git
+	branch = master
+[submodule "ogg"]
+	path = ogg
+	url = https://github.com/xiph/ogg.git
+	branch = master
```
ShFil119, line 15: if I see correctly these repos are no longer updated. 

(see gitlab for example: https://gitlab.xiph.org/xiph/opus-tools)

premake5.lua
```diff
@@ -186,10 +191,24 @@ project "re3"
...
+
 		libdirs { "milessdk/lib" }
+	libdirs { "openal-soft/libs/Win32" }
+	libdirs { "mpg123/lib" }
+	libdirs { "libsndfile/lib" }
+	if _OPTIONS["with-opus"] then
+		filter "platforms:win*"
+			libdirs { "ogg/win32/VS2015/Win32/%{cfg.buildcfg}" }
+			libdirs { "opus/win32/VS2015/Win32/%{cfg.buildcfg}" }
+			libdirs { "opusfile/win32/VS2015/Win32/Release-NoHTTP" }
```
ShFil119, line 208: it'd be possible to avoid `VS2015`? or extract as variable?

src/audio/oal/stream.cpp
```diff
@@ -170,15 +175,111 @@ class CMP3File : public IDecoder
...
+		m_nRate(0),
+		m_nChannels(0)
+	{
+		int ret;
+		m_FileH = op_open_file(path, &ret);
+
+		if (m_FileH) {
+			m_nChannels = op_head(m_FileH, 0)->channel_count;
+			m_nRate = op_head(m_FileH, 0)->input_sample_rate;
+			auto tags = op_tags(m_FileH, 0);
```
ShFil119, line 197: wow

.gitmodules
blattersturm: I'm not sure why it is taken in as submodule when static libraries are used still instead of being part of the build - did they include these .libs in the repo?

premake5.lua
Sergeanur: That's how their repo structured, so no. Although we do need a different path for Linux (I don't know what it is)

.gitmodules
Sergeanur: I don't see a problem with using a submodule.

PR #605, opened 2020-05-28
Added toggles for the TimeBars and the "DebugStuffInRelease" to the Debug Menu

rollschuh2282:
also added a Toggle for rendering Water.

### DIFF
--- a/src/core/timebars.cpp
+++ b/src/core/timebars.cpp
@@ -11,6 +11,8 @@
 // enables frame time output
 #define FRAMETIME
 
+bool gbHideTimeBars;
+
 struct sTimeBar
 {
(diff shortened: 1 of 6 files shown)


src/core/timebars.cpp
```diff
@@ -76,6 +78,9 @@ float Diag_GetFPS()
 
 void tbDisplay()
 {
+	if (gbHideTimeBars)
```
erorcun, line 81: can you please add as a comment that this is custom code? except that lgtm

aap: It would make more sense to put that if where tbDisplay is called, because it's actually never called in mobile.

PR #607, opened 2020-05-29
Added toggles for the TimeBars and the "DebugStuffInRelease" to the Debug Menu 

rollschuh2282:
also added a Don´t render Water toggle
(and changed that outdated comment in the config.h for the debug script loader

### DIFF
--- a/src/render/WaterLevel.cpp
+++ b/src/render/WaterLevel.cpp
@@ -35,6 +35,8 @@ bool CWaterLevel::WavesCalculatedThisFrame;
 RpAtomic *CWaterLevel::ms_pWavyAtomic;
 RpGeometry *CWaterLevel::apGeomArray[8];
 int16 CWaterLevel::nGeomUsed;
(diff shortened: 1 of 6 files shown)


src/render/WaterLevel.cpp
```diff
@@ -35,6 +35,8 @@ bool CWaterLevel::WavesCalculatedThisFrame;
 RpAtomic *CWaterLevel::ms_pWavyAtomic;
 RpGeometry *CWaterLevel::apGeomArray[8];
 int16 CWaterLevel::nGeomUsed;
+//"Custom" Don�t Render Water Toggle
+bool gbDontRenderWater;
```
erorcun, line 39: those are not custom, you can remove the comments :D I was talking about gbShowTimebars (and gbDontRenderWater actually but later realized that it's original)

rollschuh2282: let´s just leave them for now.
we can remove the comments later.
:D

PR #616, opened 2020-06-04
Fix headshots issue

### LINKED ISSUE #570: [Original game] npc's can't be killed under certain angles from above.
example above^
I shot the whole magazine into the mafia member from above during this rampage and i could hear his pain sounds and blood comming out of him but he didn't die till i changed an "angle" to kill him.
happens in original game as well, i also noticed it in PS2 with roof sniper-rifle headshots rampage in Shorside Vale.

ShFil119:
fixes #570


Noskillx:
lgtm in-game for me :D

Sergeanur:
Maybe make DoesLOSBulletHitPed return true instead?

aap:
yeah, that seems nicer

PR #617, opened 2020-06-05
some original VC's cheats and comparator for them

majesticCoding:


src/core/Pad.cpp
```diff
@@ -973,163 +979,173 @@ void CPad::AddToCheatString(char c)
 }
 #endif
 
+int CheatComp(char* sourceStr, char* origCheatStr)
```
aap, line 982: this is called Cheat_strncmp and even though it seems to be implemented a bit differently this code is basically ok, but add a check to return 1 if `i` is out of bounds of `cheatCodeVals`

src/core/Pad.cpp
```diff
@@ -973,163 +979,173 @@ void CPad::AddToCheatString(char c)
...
+
+	for (int32 i = 0; i < strlen(origCheatStr); i++) {
+		if (sourceStr[i] != origCheatStr[i] - cheatCodeVals[i]) {
+			return 1;
+		}
+	}
+	return 0;
+}
+
 void CPad::AddToPCCheatString(char c)
```
aap, line 994: this whole function is a big if/else chain in VC, not only ifs.

PR #626, opened 2020-06-10
more original cheats

majesticCoding:
- CSmokeTrails and CSmokeTrail classes added in new files
- more cheats added


src/peds/Ped.cpp
```diff
@@ -13423,6 +13423,7 @@ CPed::Render(void)
 
 					// TODO(Miami): What are those numbers?!
 					// localAdjMat.Rotate(flt_691FFC * flt_97F878, flt_691FF8 * flt_97F87C, flt_691FF4 * flt_97F880);
+					localAdjMat.Rotate(0.078f, -0.519f, -0.0185f);
```
aap, line 13426: define `CVector vecTestTemp(-1.0f, -1.0f, -1.0f);` externally and multiply the angles by that like the game does. also the angles were probably in degress so convert them and use DEGTORAD.

PR #632, opened 2020-06-14
expanded modelIndices, more original cheats, slightly edited CStreami…

majesticCoding:


src/modelinfo/ModelIndices.h
```diff
@@ -159,14 +159,62 @@ enum
 
 	MI_TAXI_D = 28,	// HMOCA
```
erorcun, line 160: you should move that below MI_HMOCA and make it equal to that, like
```MI_HMOCA, 
MI_TAXI_D = MI_HMOCA,```
but if you still don't know how to edit PRs then it's not problem. lgtm except that

majesticCoding: can't fix it, I couldn't make git pull with new commits today, and then decided to do git reset --hard and lost all my changes :(

PR #637, opened 2020-06-21
script logging

Nick007J:


Nick007J:
I added HANDLE types for some commands but it was too boring to mark them manually, so I left them as ints mostly

PR #639, opened 2020-06-25
Restored beta police system(disabled), fixes from miami, debug info for CFO crash

erorcun:


src/control/Phones.cpp
```diff
@@ -259,6 +294,32 @@ CPhoneInfo::SetPhoneMessage_Repeatedly(int phoneId, wchar *msg1, wchar *msg2, wc
...
+			nearestPhoneId = phoneId;
+		}
+	}
+	m_aPhones[nearestPhoneId].m_nState = PHONE_STATE_MESSAGE_REMOVED;
+
+	CPhone oldPhone = m_aPhones[into];
+	m_aPhones[into] = m_aPhones[nearestPhoneId];
+	m_aPhones[nearestPhoneId] = oldPhone;
+	m_nScriptPhonesMax++;
+	return m_nScriptPhonesMax - 1;
```
Nick007J, line 319: `return m_nScriptPhonesMax++;`?

erorcun: I don't use the return value anyway so I'll remove it

PR #641, opened 2020-07-01
CBoat and friends

aap:


src/vehicles/Floater.cpp
```diff
@@ -7,18 +7,40 @@
...
+float fBoatVolumeDistributionDinghy[9] = {
+	0.65f, 0.85f, 0.65f,
+	0.85f, 1.1f, 0.85f,
+	0.65f, 0.95f, 0.65f
+};
+float fBoatVolumeDistributionSpeed[9] = {
+	0.7f, 0.9f, 0.7f,
+	0.95f, 1.0f, 0.95f,
+	0.6f, 0.7f, 0.6f
+};
```
ShFil119, line 43: const

aap:
I don't think we need it. we didn't use it in other places either.

PR #651, opened 2020-07-06
Miami - stats cleanup,hidden packages rewards,gxt string fix

Domiiniik:


Domiiniik:
someone broke that shit, i give up.

PR #678, opened 2020-08-05
added support for FreeBSD

blingu:
Tested on 12.1-RELEASE

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -76,10 +76,6 @@ workspace "re3"
 			"linux-amd64-librw_gl3_glfw-oal",
 			"linux-arm-librw_gl3_glfw-oal",
 		}
-	filter { "system:bsd" }
-		platforms {
-			"bsd-amd64-librw_gl3_glfw-oal"
-		}
 
 	filter "configurations:Debug"
 		defines { "DEBUG" }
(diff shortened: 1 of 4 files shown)


src/core/CdStreamPosix.cpp
```diff
@@ -150,9 +150,11 @@ CdStreamInit(int32 numChannels)
         ASSERT(0);
         return;
     }
-
+#ifdef __linux__
 	_gdwCdStreamFlags = O_RDONLY | O_NOATIME;
-
+#elif __FreeBSD__
```
erorcun, line 155: you can do this just #else

src/skel/glfw/glfw.cpp
```diff
@@ -209,7 +209,11 @@ double
 psTimer(void)
 {
 	struct timespec start; 
+#ifdef __linux__
 	clock_gettime(CLOCK_MONOTONIC_RAW, &start);
+#elif __FreeBSD__
```
erorcun, line 214: same here

premake5.lua
```diff
@@ -1,308 +1,328 @@
-newoption {
```
erorcun, line 1: can you fix the formatting pls? :/

src/core/CdStreamPosix.cpp
blingu: fixed

src/skel/glfw/glfw.cpp
blingu: fixed

premake5.lua
blingu: fixed

PR #681, opened 2020-08-07
 remove CoUninitialize calls for FIX_BUGS (fixes PIX crash) and don't …

withmorten:
…check return value for CoInitialize


withmorten:
aap noticed that PIX crashes when trying to do a frame capture ... this should fix that - the problem? Never call CoUninitialize without having actually called CoInitialize first! Solution: Don't call CoUninitialize, ever!

PR #695, opened 2020-08-24
new frontend customization

erorcun:
- if GRAPHICS_MENU_OPTIONS isn't defined Display menu will accept 2 more options, if it's more then that it will create Advanced Display menu and add a shortcut to Display
- All repositionings within GRAPHICS_MENU_OPTIONS has been made according to Ash's proposal
- Island loading slider disabled until someone port it to CFO
- Now you can add new screen with FrontendScreenAdd and it will return the new page ID
- FrontendScreenSetPosition is renamed to FrontendScreenSetCursor;
    - and has a new parameter: overwrite. So you can both append or overwrite options at any position. (and it will still progress the cursor)
    - negative positions now really start counting from bottom, no hardcoded back button detection thing. but -1 means the very bottom, -2 means one above of it etc.
- Live reload will also undo overwrites/replacements. Read explanations at the top of frontendoption.h if you wonder what is live reload.
- DrawFunc now has a new parameter: `userHovering`
- Static/select options how also handle HelperTexts.
- FrontendOptionAddDynamic/FrontendOptionAddStatic now accept `save` parameter, and can save options to gta3.set itself.


aap:
looks ok at first glace. will have to try it out...

PR #698, opened 2020-08-30
cheats, fixes, cosmetic changes

majesticCoding:


src/core/Frontend.cpp
```diff
@@ -5369,6 +5369,7 @@ CMenuManager::ConstructStatLine(int rowIdx)
 	STAT_LINE("FEST_CC", &CStats::CriminalsCaught, false, nil);
 	STAT_LINE("FEST_FE", &CStats::FiresExtinguished, false, nil);
 	STAT_LINE("DAYPLC", &(nTemp = CTimer::GetTimeInMilliseconds() + 100), false, nil);
+	//TODO(MIAMI): move this function to the CStats and add reading of Stat lines tied with "MEDIA" for the "CHASESTAT" cheatcode
```
erorcun, line 5372: this function is hella TODO so this comment is redundant actually, but not a problem ofc

src/render/Hud.cpp
```diff
@@ -524,6 +526,54 @@ void CHud::Draw()
 			}
 		}
 
+		if (CStats::ShowChaseStatOnScreen != 0) {
+			static int32 nMediaLevelCounter;
```
erorcun, line 530: this should be initialized to 0, before the if block

src/render/Hud.cpp
```diff
@@ -524,6 +526,54 @@ void CHud::Draw()
...
+
+				CRGBA colour;
+				if (CTimer::GetTimeInMilliseconds() & 0x200)
+					colour = CRGBA(204, 0, 185, 180);
+				else
+					colour = CRGBA(178, 0, 162, 180);
+				CFont::SetColor(colour);
+				CFont::PrintString(SCREEN_SCALE_FROM_RIGHT(110.0f), SCREEN_SCALE_Y(113.0f), gUString);
+				
+				static wchar *prevChaseString;
```
erorcun, line 563: this is not a static

src/render/Hud.cpp
```diff
@@ -524,6 +526,54 @@ void CHud::Draw()
 			}
 		}
 
+		if (CStats::ShowChaseStatOnScreen != 0) {
+			static int32 nMediaLevelCounter;
+			float fCurAttentionLevel = CWorld::Players[CWorld::PlayerInFocus].m_fMediaAttention;
+			if (HUD_TEXT_SCALE_X * CStats::HighestChaseValue > fCurAttentionLevel
```
erorcun, line 532: looks like 0.7 used in here is coincidentally same number with HUD_TEXT_SCALE_X, you can use hardcoded 0.7 in here.

src/render/Hud.cpp
majesticCoding: fixed

src/render/Hud.cpp
majesticCoding: fixed

erorcun (approved):

src/core/Pad.cpp
```diff
@@ -208,15 +209,20 @@ void HealthCheat()
 	}
 }
 
-void VehicleCheat(bool something, int model)
+void VehicleCheat(int model)
 {
 	CHud::SetHelpMessage(TheText.Get("CHEAT1"), true);
-	CStreaming::RequestModel(model, 0);
-	CStreaming::LoadAllRequestedModels(something);
+	CStreaming::RequestModel(model, 1);
```
aap, line 215: use the enum here

PR #711, opened 2020-09-25
remove version

hamed34434:
plz remove version 1.0 in screen


erorcun:
We won't do it, you may ask to Switch fork. And this isn't what PRs are for.

PR #716, opened 2020-09-26
How to compile gta 3 in Linux?

COERSA-bra:
I'm looking to compile gta 3 on Linux, but I don't know how to do it


ShFil119:
https://github.com/GTAmodding/re3/wiki/Building-on-Linux

PR #722, opened 2020-09-28
fix for frequency miscalculation

theR4K:
in mss freq can become negative bcs max used on unsigned values.


src/audio/AudioManager.cpp
```diff
@@ -794,7 +794,7 @@ cAudioManager::ProcessActiveQueues()
 							if (sample.m_nFrequency != m_asActiveSamples[j].m_nFrequency) {
 								int32 freq;
 								if (sample.m_nFrequency <= m_asActiveSamples[j].m_nFrequency) {
-									freq = Max(sample.m_nFrequency, m_asActiveSamples[j].m_nFrequency - 6000);
+									freq = Max((int32)sample.m_nFrequency, (int32)(m_asActiveSamples[j].m_nFrequency - 6000));
```
ShFil119, line 797: can you change to?
`freq = Max((int32)sample.m_nFrequency, (int32)(m_asActiveSamples[j].m_nFrequency) - 6000);`

PR #723, opened 2020-09-28
Fixes for Wayland (HiDPI and mouse lock) support, FreeBSD

myfreeweb:
This PR:

- Fixes HiDPI support on Wayland (will fix on macOS too when someone does macOS support) by using framebuffer size, not window size (the latter is in logical pixels, not real GL pixels)
  - merge together with librw PR: https://github.com/aap/librw/pull/35
- Fixes mouse lock on Wayland (you can't just warp the mouse to the center, let GLFW handle it)
- Adds more architectures to `bsd` build (I'll test on arm64 soon)
- Adds `CLOCK_MONOTONIC_FAST` usage
- needs testing on other platforms to make sure nothing is broken of course :)

### DIFF
--- a/src/skel/glfw/glfw.cpp
+++ b/src/skel/glfw/glfw.cpp
@@ -1397,8 +1397,11 @@ _InputTranslateShiftKeyUpDown(RsKeyCodes *rs) {
 // TODO this only works in frontend(and luckily only frontend use this). Fun fact: if I get pos manually in game, glfw reports that it's > 32000
 void
 cursorCB(GLFWwindow* window, double xpos, double ypos) {
-	FrontEndMenuManager.m_nMouseTempPosX = xpos;
-	FrontEndMenuManager.m_nMouseTempPosY = ypos;
+	int bufw, bufh, winw, winh;
+	glfwGetWindowSize(window, &winw, &winh);
+	glfwGetFramebufferSize(window, &bufw, &bufh);
+	FrontEndMenuManager.m_nMouseTempPosX = xpos * (bufw / winw);
+	FrontEndMenuManager.m_nMouseTempPosY = ypos * (bufh / winh);
 }
 
 void
--- a/src/skel/glfw/glfw.cpp
+++ b/src/skel/glfw/glfw.cpp
@@ -1628,6 +1628,8 @@ main(int argc, char *argv[])
 #endif
 		{
 			glfwPollEvents();
+			glfwSetInputMode(PSGLOBAL(window), GLFW_CURSOR,
+			                 (FrontEndMenuManager.m_bMenuActive && !PSGLOBAL(fullScreen)) ? GLFW_CURSOR_HIDDEN : GLFW_CURSOR_DISABLED);
 			if( ForegroundApp )
 			{
 				switch ( gGameState )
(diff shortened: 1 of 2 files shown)


erorcun:
Thanks, I'll be able to try and review it in a few days.

src/skel/glfw/glfw.cpp
```diff
@@ -1397,8 +1399,11 @@ _InputTranslateShiftKeyUpDown(RsKeyCodes *rs) {
 // TODO this only works in frontend(and luckily only frontend use this). Fun fact: if I get pos manually in game, glfw reports that it's > 32000
 void
 cursorCB(GLFWwindow* window, double xpos, double ypos) {
-	FrontEndMenuManager.m_nMouseTempPosX = xpos;
-	FrontEndMenuManager.m_nMouseTempPosY = ypos;
+	int bufw, bufh, winw, winh;
+	glfwGetWindowSize(window, &winw, &winh);
+	glfwGetFramebufferSize(window, &bufw, &bufh);
+	FrontEndMenuManager.m_nMouseTempPosX = xpos * (bufw / winw);
+	FrontEndMenuManager.m_nMouseTempPosY = ypos * (bufh / winh);
```
erorcun, line 1406: This CB only used for Frontend(menus) as stated above the function. Mouse capturing while playing is done in CPad::UpdateMouse with glfwGetCursorPos, did you miss it or does glfwGetCursorPos work correct for HiDPI?

src/skel/glfw/glfw.cpp
```diff
@@ -1625,6 +1630,8 @@ main(int argc, char *argv[])
 #endif
 		{
 			glfwPollEvents();
+			glfwSetInputMode(PSGLOBAL(window), GLFW_CURSOR,
+			                 (FrontEndMenuManager.m_bMenuActive && !PSGLOBAL(fullScreen)) ? GLFW_CURSOR_HIDDEN : GLFW_CURSOR_DISABLED);
```
erorcun, line 1634: there is a few problems in here:
- why set input mode on every frame?
- I agree with you on using GLFW_CURSOR_DISABLED, but you should wrap our mouse centering with `#ifndef RW_GL3` in that case. it's in `main.cpp`, should show up when you search for `// This is from SA, but it's nice for windowed mode`. 
- `fullScreen` isn't updated after screen mode change, you'll have to update it if you want to use that variable.

src/skel/glfw/glfw.cpp
myfreeweb: While playing, we don't have a cursor, it's all relative movement of the camera — so the only thing that could be incorrect is the camera movement speed (~mouse sensitivity). I guess it might be a bit high, but it was playable.

src/skel/glfw/glfw.cpp
myfreeweb: > why set input mode on every frame?

I could not find the places where this condition would change :(

erorcun: at the end of `CMenuManager::SwitchMenuOnAndOff()` you will see `if (m_bMenuActive != menuWasActive)`, you can add your code under it. you can use `!m_nPrefsWindowed` for checking if it's fullscreen. also don't forget to wrap it with `#ifdef RW_GL3` if it's certain that you want to add it there.

erorcun:
I'll merge and fix it myself

Issue #730, opened 2020-09-29
Preliminary non-working macOS support


Tatsh:
Makes the game show the menu but it freezes and hangs when trying to start a game.

Patches and then manual editing of `re3.Makefile`:

```patch
diff --git a/premake5.lua b/premake5.lua
index 881035ef..37901496 100644
--- a/premake5.lua
+++ b/premake5.lua
@@ -80,6 +80,11 @@ workspace "re3"
 			"linux-arm64-librw_gl3_glfw-oal",
 		}
 
+	filter { "system:macosx" }
+	    platforms {
+	        "macosx-arm64-librw_gl3_glfw-oal",
+	    }
+
 	filter { "system:bsd" }
 		platforms {
 			"bsd-amd64-librw_gl3_glfw-oal"
@@ -101,6 +106,12 @@ workspace "re3"
 	filter { "platforms:bsd*" }
 		system "bsd"
 
+	filter { "platforms:mac*" }
+	    system "macosx"
+
+	filter { "platforms:darwin" }
+	    system "macosx"
+
 	filter { "platforms:*x86*" }
 		architecture "x86"
 
diff --git a/src/skel/glfw/glfw.cpp b/src/skel/glfw/glfw.cpp
index b9dbf5ac..93f73a11 100644
--- a/src/skel/glfw/glfw.cpp
+++ b/src/skel/glfw/glfw.cpp
@@ -81,7 +81,9 @@ DWORD _dwOperatingSystemVersion;
 #include "resource.h"
 #else
 long _dwOperatingSystemVersion;
+#ifndef __APPLE__
 #include <sys/sysinfo.h>
+#endif
 #include <stddef.h>
 #include <locale.h>
 #include <signal.h>
@@ -445,14 +447,22 @@ psInitialize(void)
 #endif
 
 #endif
+#ifndef __APPLE__
 	struct sysinfo systemInfo;
 	sysinfo(&systemInfo);
+#endif
 
+#ifndef __APPLE__
 	_dwMemAvailPhys = systemInfo.freeram;
+#else
+    _dwMemAvailPhys = 230535168; // FIXME get the real value from sysctlbyname(), vm.pageout_* variables, multiply by page size 4096
+#endif
 	_dwOperatingSystemVersion = OS_WINXP; // To fool other classes
 
+#ifndef __APPLE__
 	debug("Physical memory size %u\n", systemInfo.totalram);
 	debug("Available physical memory %u\n", systemInfo.freeram);
+#endif
 
[...]

PR #734, opened 2020-09-30
macOS support

Tatsh:
Based on https://github.com/GTAmodding/re3/issues/730, mostly @MrYadro's work.

~~The game hangs on boot. Needs a patch for sysinfo in `src/skel/glfw/glfw.cpp` to get memory information.~~

~~Best is to use [sysctlbyname(3)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/sysctlbyname.3.html) and link against libSystem (`-lSystem`). This is different on macOS, as what's considered free memory is like so:~~

~~FreeMemory = TotalMemory - AppMemory - WiredMemory - CompressedMemory~~

~~I would say it's not that important to put a real value there, is it?~~

### DIFF
--- a/src/core/CdStreamPosix.cpp
+++ b/src/core/CdStreamPosix.cpp
@@ -4,6 +4,7 @@
 #include <pthread.h>
 #include <signal.h>
 #include <semaphore.h>
+#include <stdlib.h>
 #include <sys/types.h>
 #include <unistd.h>
 #include <sys/time.h>
@@ -21,6 +22,30 @@
 #define CDDEBUG(f, ...)   debug ("%s: " f "\n", "cdvd_stream", ## __VA_ARGS__)
 #define CDTRACE(f, ...)   printf("%s: " f "\n", "cdvd_stream", ## __VA_ARGS__)
 
+#ifdef __APPLE__
+#define COMPAT_SEM_T sem_t *
+int compat_sem_init(sem_t **ptr, __attribute__((unused)) int x, __attribute__((unused)) int y) {
+  *ptr = sem_open("/semaphore", O_CREAT, 0644, 1);
+  return *ptr == SEM_FAILED ? -1 : 0;
+}
+int compat_sem_post(sem_t **sem) {
+  return sem_post(*sem);
+}
+int compat_sem_wait(sem_t **sem) {
+  return sem_wait(*sem);
+}
+int compat_sem_destroy(sem_t **sem, const char * name) {
+  sem_close(*sem);
+  sem_unlink(name);
+}
+#else
+#define COMPAT_SEM_T sem_t
+#define compat_sem_post sem_post
+#define compat_sem_destroy(x, y) sem_destroy(x)
+#define compat_sem_init sem_init
+#define compat_sem_wait sem_wait
+#endif
+
 // #define ONE_THREAD_PER_CHANNEL // Don't use if you're not on SSD/Flash. (Also you may want to benefit from this via using all channels in Streaming.cpp)
 
 bool flushStream[MAX_CDCHANNELS];
@@ -35,11 +60,11 @@ struct CdReadInfo
 	int32 nStatus;
 #ifdef ONE_THREAD_PER_CHANNEL
 	int8 nThreadStatus; // 0: created 1:initalized 2:abort now
-    pthread_t pChannelThread;
-    sem_t pStartSemaphore;
+  pthread_t pChannelThread;
+  COMPAT_SEM_T pStartSemaphore;
 #endif
-	sem_t pDoneSemaphore; // used for CdStreamSync
-	int32 hFile;
+  COMPAT_SEM_T pDoneSemaphore; // used for CdStreamSync
+  int32 hFile;
 };
 
 char gCdImageNames[MAX_CDIMAGES+1][64];
@@ -51,7 +76,7 @@ char *gImgNames[MAX_CDIMAGES];
 
 #ifndef ONE_THREAD_PER_CHANNEL
 pthread_t _gCdStreamThread;
-sem_t gCdStreamSema; // released when we have new thing to read(so channel is set)
+COMPAT_SEM_T gCdStreamSema; // released when we have new thing to read(so channel is set)
 int8 gCdStreamThreadStatus; // 0: created 1:initalized 2:abort now
 Queue gChannelRequestQ;
 bool _gbCdStreamOverlapped;
@@ -76,7 +101,7 @@ CdStreamInitThread(void)
 	gChannelRequestQ.tail = 0;
 	gChannelRequestQ.size = gNumChannels + 1;
 	ASSERT(gChannelRequestQ.items != nil );
-    status = sem_init(&gCdStreamSema, 0, 0);
+    status = compat_sem_init(&gCdStreamSema, 0, 0);
 #endif
 
 
@@ -91,7 +116,7 @@ CdStreamInitThread(void)
 	{
 		for ( int32 i = 0; i < gNumChannels; i++ )
 		{
-		    status = sem_init(&gpReadInfo[i].pDoneSemaphore, 0, 0);
+		    status = compat_sem_init(&gpReadInfo[i].pDoneSemaphore, 0, 0);
 
 			if (status == -1)
 			{
@@ -100,7 +125,7 @@ CdStreamInitThread(void)
 				return;
 			}
 #ifdef ONE_THREAD_PER_CHANNEL
-		    status = sem_init(&gpReadInfo[i].pStartSemaphore, 0, 0);
+		    status = compat_sem_init(&gpReadInfo[i].pStartSemaphore, 0, 0);
 
 			if (status == -1)
 			{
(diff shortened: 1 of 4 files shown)


Tatsh:
I think this is the error causing the hanging:

```
[DBG]: Streaming buffer size is 920 sectors[DBG]: Memory allocated to Streaming is 78MBSignal Sema Error
```

`src/core/Streaming.cpp:201` is missing an ending `\n` and one of these 3 print:

```
src/core/CdStream.cpp:226:                      printf("Signal Sema Error\n");
src/core/CdStreamPosix.cpp:258:        printf("Signal Sema Error\n");
src/core/CdStreamPosix.cpp:261:        printf("Signal Sema Error\n");
```

The above prints, and spindump looks like it's waiting on a lock:

```
  Thread 0x1a1085    DispatchQueue "com.apple.main-thread"(1)    1000 samples (1-1000)    priority 31 (base 31)    cpu time 9.862s (34.4G cycles, 17.6G instructions, 1.95c/i)
  1000  start + 1 (libdyld.dylib + 109769) [0x7fff6cf92cc9]
    1000  main + 1780 (glfw.cpp:1829,7 in re3 + 2049924) [0x10cfd9784]
      1000  CGame::Initialise(char const*) + 597 (Game.cpp:391,2 in re3 + 789765) [0x10cea5d05]
        962  CStreaming::LoadAllRequestedModels(bool) + 445 (Streaming.cpp:1859,10 in re3 + 1778701) [0x10cf9740d]
          910  sem_wait + 10 (libsystem_kernel.dylib + 137966) [0x7fff6d0f4aee]
           *344  hndl_unix_scall64 + 10 (kernel + 791034) [0xffffff80002c11fa] (running)
```

[Spindump.txt](https://github.com/GTAmodding/re3/files/5303671/Spindump.txt)

MrYadro:
I think you could change gcc (10) for gcc-8 at mac build script. Because I have same errors with gcc (10)

Tatsh:
Main difference I noticed was I got errors when I set CC to gcc even though it's an alias for Clang. I set the compiler to Clang and that removed some of the errors but I'm not sure what's causing the current error. Maybe include path order which I don't know how to affect in Premake.

Tatsh:
Basically setting CC to gcc (as Travis does by default) seems to turn on a slightly different C++ standard or implicit flags.

MrYadro:
setting CXXFLAGS=-std=c++11 fixed those errors for me!

Tatsh:
> setting CXXFLAGS=-std=c++11 fixed those errors for me with clang!

Thanks for the tip. I changed the project to use `-std=gnu++14` for macOS builds with Clang.

MrYadro:
I just made MacOS version to work, all you have to do is follow this instruction https://heldercorreia.com/semaphores-in-mac-os-x-fd7a7418e13b

Tatsh:
A macro that gets defined conditionally is the best way. The macro would add & on non-Apple before the argument.

Tatsh:
Thanks for figuring out that issue. I made the macros. This works great now.

@Nick007J @Sergeanur @erorcun and others, this is ready for review.

Tatsh:
Still getting some crashes after skipping the intro. Built on another MacBook Pro to confirm it wasn't just my machine. Also deleted `userfiles` before launching. Spindump attached.

Could you test my latest version @MrYadro ? Maybe I missed something when I wrote to use the macros. If not, then it's files corrupted somehow:

```
../vendor/librw/src/geometry.cpp:183: Couldn't find chunk MATLIST
[DBG]: Failed to load LODNcoast1
```

Right before the hang ^

[Spindump2.txt.gz](https://github.com/GTAmodding/re3/files/5308985/Spindump2.txt.gz)

Tatsh:
Still, the game shouldn't hang like this anywhere. There should be a watchdog timer for these locks that will time out and exit or go back to the menu, hopefully with a message.

MrYadro:
I'll look into it in the evening, but for now I see that you are creating semaphores with same name, but destroying with different. Not sure if name matter at all.

Tatsh:
I switched to named semaphores for all POSIX platforms, as this will be much easier to maintain. This fixes the issues I had.

Tatsh:
Found a new crash point. Going to Controller setup -> Redefine controls crashes:

```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x000000210c701ec0
Exception Note:        EXC_CORPSE_NOTIFY

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   re3                           	0x000000010c47cd9e CControllerConfigManager::GetControllerSettingTextWithOrderNumber(e_ControllerAction, eContSetOrder) + 14 (ControllerConfig.cpp:1899)
1   re3                           	0x000000010c4bbddc CMenuManager::DrawControllerBound(int, int, int, signed char) + 1244 (Frontend.cpp:1860)
2   re3                           	0x000000010c4bd74a CMenuManager::DrawControllerSetupScreen() + 3338 (Frontend.cpp:2195)
3   re3                           	0x000000010c4be22c CMenuManager::DrawFrontEndNormal() + 1532
4   re3                           	0x000000010c4bdc1b CMenuManager::DrawFrontEnd() + 107 (Frontend.cpp:2280)
5   re3                           	0x000000010c5fb386 FrontendIdle() + 230 (main.cpp:1249)
6   re3                           	0x000000010c5fb525 AppEventHandler + 325
7   re3                           	0x000000010c602b21 RsEventHandler + 17 (skeleton.cpp:218)
8   re3                           	0x000000010c5f89bb main + 2043 (glfw.cpp:1782)
9   libdyld.dylib                 	0x00007fff6cf92cc9 start + 1
```

src/core/CdStreamPosix.cpp
```diff
@@ -205,7 +205,7 @@ GetGTA3ImgSize(void)
         return 0;
     }
 ok:
-	return statbuf.st_size;
+	return (uint32)statbuf.st_size;
```
erorcun, line 208: why limit gta3.img to 4 GB? except that looks good to me btw

src/core/CdStreamPosix.cpp
```diff
@@ -91,18 +91,18 @@ CdStreamInitThread(void)
 	{
 		for ( int32 i = 0; i < gNumChannels; i++ )
 		{
-		    status = sem_init(&gpReadInfo[i].pDoneSemaphore, 0, 0);
+      gpReadInfo[i].pDoneSemaphore = sem_open("/semaphore_done", O_CREAT, 0644, 1);
```
erorcun, line 94: for every channel (i) there is gonna be a new thread, so all of them shouldn't have the same semaphore name i guess? this is same for other `semaphore_done`s and `semaphore_start`s below

Tatsh: I was thinking that too, but `sem_open` and `sem_unlink` functions are supposed to increase and decrease the reference counts for the names rather than overwrite. These should be getting cleared on `sem_unlink` when the reference count reaches 0.

From the Linux man page for `sem_open(3)`:

>The oflag argument specifies flags that control the operation of the call. If O_CREAT is specified  in  oflag, then the semaphore is created if it does not already exist.

From `sem_ulink(3)`:

>sem_unlink()  removes  the  named semaphore referred to by name.  The semaphore name is removed immediately.  The semaphore is destroyed once all other processes that have the semaphore open close it.

(I know it's kind of confusing and the term *reference count* isn't used here. Process can be a thread or an actual process.)

In BSD's manual (Darwin version) this is much more clear (`sem_unlink(2)`):

>The named semaphore named name is removed. If the semaphore is in use by other processes, then name is immediately disassociated with the semaphore, but the semaphore itself will not be removed until all references to it have been closed.

(I think they meant to write the specific semaphore in the scope of your thread/process, not others.)

> Subsequent calls to sem_open() using name will refer to or create a new semaphore named name.

Tatsh: Implementation of `sem_close` checks a `refcnt` field in Glibc: https://github.com/bminor/glibc/blob/master/nptl/sem_close.c#L60

Tatsh:
Also if you think the CI for macOS is too slow, I can comment it out of `travis.yml` until maybe this situation from Travis improves.

https://github.com/travis-ci/travis-ci/issues/7304

It still is slow, but perhaps not as slow as when this issue was filed.

src/core/CdStreamPosix.cpp
erorcun: ooh ok then, thanks.

erorcun:
merging it, thank you :)
4 min. as build time doesn't seem too much to me right now, but if it will tend to take longer then we can consider it.

erorcun:
just to let you know, today I checked the addresses of semaphores with same name after I experienced a deadlock, and they had same addresses. at least on linux.

I'll fix it later today.

Tatsh:
Just curious, did this deadlpck happen when loading a cutscene?

I've been doing the island skipping glitch with the tunnel near the junkyard during my testing which makes the game load a lot of stuff if not almost everything from the img files. Did not have a deadlock yet from this. Also I tested this way on Linux.

Tatsh:
I haven't tried with that defined before. I do have SSDs on everything but is it really that much faster?

erorcun:
no, it's just a starting point for future things, my latest PR(which also changes the semaphore names) makes use of it with enabling LoadAllRequestedModels to load max. 4 models at once for example. we're not processing them multi-threaded yet, but if we'll ever do it in future then seperate channels will come handy too.

I just wanted to correct that sentence: `but sem_open and sem_close functions are supposed to increase and decrease the reference counts for the names rather than overwrite.` if I got it right you must have meant it doesn't overwrite existing semaphores with same name, but it unfortunately does. since it was always single threaded without that define we never noticed it, so there was no problem without `ONE_THREAD_PER_CHANNEL` actually.

PR #735, opened 2020-09-30
car fly cheat enhancement

theR4K:
IWANTTOMASTERDODO now allows cars fly like seaplane, if comeflywithme activated. For this I recalculate fThrust and fThrustFallOff using car characteristic.
Also if IWANTTOMASTERDODO activated seaplane accelerate faster. But it's max speed stay same.


theR4K:
so now it independent cheat with name "carsareplane". Also I add some multiplier to cars engine power, bcs without it cars was much more friction(or something like) in air. For huge and slow cars I prefer seaplane values.
In general now car fly with it's max speed, or with max seaplane speed(if own specifications too low).

erorcun:
that wasn't I meant, I meant you should change/alter the behaviour of COMEFLYWITHME/AllCarsAreDodo with `BETTER_ALLCARSAREDODO_CHEAT`.

What you did right now is also lgtm but we can't name it `BETTER_ALLCARSAREDODO_CHEAT` anymore, you should name it something like `CARSAREPLANE_CHEAT` etc.

erorcun (approved):

ShFil119 (approved):
also looks ok

PR #736, opened 2020-09-30
Readme Adjustments

Krutonium:
I made a couple of changes to your Readme:
 - First, the coding style is now its own document (And added instructions to read it)
 - Second, I suggested a better way to clone that removes steps, specifically you can use `--recursive` to skip having to `init` and `update` the submodules.

I would also have tried to update the discord link, but I was told in an issue earlier that it was intentionally broken at the moment. (I suspect because of the Switch port?)

Questions, Comments, Edit Requests, Send them and I'll try to address them.

### DIFF
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -1,0 +1,107 @@
+# Coding style
+
+I started writing in [Plan 9 style](http://man.cat-v.org/plan_9/6/style),
+but realize that this is not the most popular style, so I'm willing to compromise.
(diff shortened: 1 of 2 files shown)


Krutonium:
Unrelated, would you like me to make Travis-ci work properly? I notice right now it's throwing out that the build is bad because of apt not liking a repo.

erorcun:
looks like travis works right now, thanks.
merged

PR #740, opened 2020-10-02
Provide support for big IMG

ShFil119:
Not sure if needed to wrap it with some defines. 
Edit: Done.

### DIFF
--- a/src/core/CdStream.cpp
+++ b/src/core/CdStream.cpp
@@ -242,8 +242,15 @@ CdStreamRead(int32 channel, void *buffer, uint32 offset, uint32 size)
 		else
 			return STREAM_SUCCESS;
 	}
-	
+
+#ifdef BIG_IMG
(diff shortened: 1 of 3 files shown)


src/core/CdStream.cpp
```diff
@@ -242,9 +242,12 @@ CdStreamRead(int32 channel, void *buffer, uint32 offset, uint32 size)
 		else
 			return STREAM_SUCCESS;
 	}
-	
-	SetFilePointer(hImage, _GET_OFFSET(offset) * CDSTREAM_SECTOR_SIZE, nil, FILE_BEGIN);
-	
+
+	LARGE_INTEGER liDistanceToMove;
```
erorcun, line 246: well, this is not original code so you should wrap it with sth like BIG_IMG_SUPPORT.
also at least you need to change return type of GetGTA3imgSize.

Nick007J (approved):
lgtm but would be better under define indeed

PR #743, opened 2020-10-04
Fix some sanitizer errors, improve POSIX streamer

erorcun:
- Now on flushing streaming will flush streaming for real by interrupting the file read
- Starting file read on a streaming channel will interrupt ongoing one as the how it was supposed to be from the beginning
- Multiple sources say fgetc and fputc accepts/outputs unsigned char (sanitizer also whines about it), therefore unsigned char casts.

If ONE_THREAD_PER_CHANNEL is defined;
- LoadAllRequestedModels will be able to read 4 files at once (not processing tho)
- Fix deadlocks introduced by Mac support


ShFil119 (approved):
lgtm

PR #745, opened 2020-10-05
VS Code settings for code and debugging (Linux/macOS x64 only)

Tatsh:
Makes it possible to click _Run_ 🡢  _Start Debugging_ in VS Code, on Linux x64 only for now. GDB only.

It really works and makes working on the code much easier, as long as you have two monitors since I don't know how to make it not run full screen.


erorcun:
can you add Release configuration?

Tatsh:
Will do, and also will add the macOS config.

Tatsh:
> can you add Release configuration?

Done. Ready to be merged.

PR #746, opened 2020-10-05
Linux/macOS: allow storage of game data in XDG standard path

Tatsh:
[XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

In `src/core/config.h`, enable `XDG_ROOT`.

Store the game data in `~/.local/share/re3`. This directory will be created if it does not exist. The `re3` binary no longer has to be in the same place. The `re3` binary won't care where it is located.

Of course, this is only for Linux/macOS but for macOS if I get unlazy I will make the storage point in a more normal spot on that OS. I suppose this could work on Windows under MSYS or Cygwin, or even WSL.

### DIFF
--- a/src/core/CdStreamPosix.cpp
+++ b/src/core/CdStreamPosix.cpp
@@ -17,6 +17,9 @@
 #include "CdStream.h"
 #include "rwcore.h"
 #include "RwHelper.h"
+#ifdef XDG_ROOT
+#include "FileMgr.h"
+#endif
 
 #define CDDEBUG(f, ...)   debug ("%s: " f "\n", "cdvd_stream", ## __VA_ARGS__)
 #define CDTRACE(f, ...)   printf("%s: " f "\n", "cdvd_stream", ## __VA_ARGS__)
@@ -139,12 +142,23 @@ CdStreamInitThread(void)
 #endif
 }
 
+static const char *gta3ImgPath = "models/gta3.img";
+
 void
 CdStreamInit(int32 numChannels)
 {
     struct statvfs fsInfo;
+    char imgPath[128] = {'\0'};
 
-    if((statvfs("models/gta3.img", &fsInfo)) < 0)
+#ifdef XDG_ROOT
+    const char *rootDir = CFileMgr::GetRootDirName();
+    strncpy(imgPath, rootDir, strlen(rootDir) - 1);
+    strcat(imgPath, "/");
+    strcat(imgPath, gta3ImgPath);
+#else
+    strcpy(imgPath, gta3ImgPath);
+#endif
+    if((statvfs(imgPath, &fsInfo)) < 0)
     {
         CDTRACE("can't get filesystem info");
         ASSERT(0);
--- a/src/core/FileMgr.cpp
+++ b/src/core/FileMgr.cpp
@@ -197,11 +197,45 @@ myfeof(int fd)
 char CFileMgr::ms_rootDirName[128] = {'\0'};
 char CFileMgr::ms_dirName[128];
 
+#ifdef XDG_ROOT
+void CFileMgr::GetHomeDirectory(char *homeDir) {
+  char **p = environ;
+  while (*p++)
+  {
+    if (!strncmp("HOME=", *p, 5)) {
+      strcpy(homeDir, *p + 5);
+      break;
+    }
+  }
+}
+#endif
+
 void
 CFileMgr::Initialise(void)
 {
+#ifdef XDG_ROOT
+  char homeDir[255];
+  bzero(homeDir, 255);
+  GetHomeDirectory(homeDir);
+  // Build up ${HOME}/.local/share/re3
+  strcat(homeDir, "/.local");
+  struct stat buf;
+  if (stat(homeDir, &buf) < 0) {
+    assert(mkdir(homeDir, 0755) == 0);
+  }
+  strcat(homeDir, "/share");
+  if (stat(homeDir, &buf) < 0) {
+    assert(mkdir(homeDir, 0755) == 0);
+  }
+  strcat(homeDir, "/re3");
+  if (stat(homeDir, &buf) < 0) {
+    assert(mkdir(homeDir, 0755) == 0);
+  }
+  strcpy(ms_rootDirName, homeDir);
+#else
 	_getcwd(ms_rootDirName, 128);
-	strcat(ms_rootDirName, "\\");
+#endif
+  strcat(ms_rootDirName, "\\");
 }
 
 void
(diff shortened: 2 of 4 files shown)


Tatsh:
Updated title, as macOS needs special code to read environment variables.

Tatsh:
As soon as this is merged, I have the macOS version.

src/core/CdStreamPosix.cpp
```diff
@@ -139,12 +142,23 @@ CdStreamInitThread(void)
...
-    if((statvfs("models/gta3.img", &fsInfo)) < 0)
+#ifdef XDG_ROOT
+    const char *rootDir = CFileMgr::GetRootDirName();
+    strncpy(imgPath, rootDir, strlen(rootDir) - 1);
+    strcat(imgPath, "/");
+    strcat(imgPath, gta3ImgPath);
+#else
+    strcpy(imgPath, gta3ImgPath);
+#endif
+    if((statvfs(imgPath, &fsInfo)) < 0)
```
erorcun, line 161: I think you should also patch CdStreamAddImage

Tatsh: Will do.

Tatsh: Are you sure this needs to be patched? Prior to any call to `CdStreamAddImage`, there's a call to `CFileMgr::ChangeDir()` (`chdir`) which, with this patch, changes to the XDG path e.g. `~/.local/share/re3`. As a result, the relative paths passed in `CdStreamAddImage` resolve fine as they will go through `casepath` to fix the backslashes and casing.

erorcun: I didn't see any? sure, coincidentally current path may be same with the one we want, but that doesn't guarantee it will always be same. in fact, you can see ChangeDir/SetDir's before all CFileMgr file reading operations, but there is none to be found before any CdStreamAddImage, so looks like it's coincidental.

Tatsh: That would mean CdStreamAddImage should check if the casepath returns nothing then try again with the XDG path prepended?

erorcun: I'd suggest just SetDir to XDG root path, but then I realized you didn't patch SetDir/ChangeDir at all... you should fix this, you can find various `fopen`s in audio files.

Tatsh: I made the change to call `SetDir("")` before doing any processing in `CdStreamAddPosix`.

I don't think `SetDir` or `ChangeDir` have to be changed as they both have a way to go to root and then prepend `ms_rootDirName` (XDG root) before adding the `dir` argument passed in. `SetDir` always does this, and `ChangeDir` only does this if the first character of the `dir` argument is `\`. All calls to `ChangeDir` begin with `\` so none are relative.

Would you consider using only one variant of these two `SetDir`/`ChangeDir` functions? My understanding is that `ChangeDir` works as if the game data directory is the root of a file system and accepts `"\\"` to mean to start at the root and not use a relative path. `SetDir` accepts only relative paths, with empty string meaning *go to root*.

It seems to me that these `SetDir` calls could be changed to 'full' paths for `ChangeDir` prepending a `\\` to the path arguments.

There are only a few unique paths used when calling either function. They all seem to want to start from the game data root.

```
CFileMgr::SetDir("");
CFileMgr::SetDir("data");
CFileMgr::SetDir("DATA");
CFileMgr::SetDir("data\\paths");
CFileMgr::SetDir("TEXT");
CFileMgr::SetDir("TEXTURES");
CFileMgr::SetDir("TXD\\");
```

Not sure what this call to `SetDir("TEXTURES")` is for. I do not have this directory anywhere in my game data extracted off the installation discs.

```
CFileMgr::ChangeDir("\\");
CFileMgr::ChangeDir("\\DATA\\");
CFileMgr::ChangeDir("\\DATA\\MAPS\\");
```

erorcun: > I don't think SetDir or ChangeDir have to be changed as they both have a way to go to root and then prepend ms_rootDirName (XDG root) before adding the dir argument passed in.

I mean if i'm not wrong currently no `chdir` sets directory to XDG path at all (which was something SetDir/ChangeDir doing). Then how `fopen`s will access to XDG path?

Tatsh: Correct, there is no chdir call to set the path to the XDG path. So, it is possible for an `fopen` call to fail if for example, `cSampleManager::InitialiseSampleBanks()` (opens `audio/sfx.*`) was called without a call to `CFileMgr::SetDir("")` / `CFileMgr::SetDir("\\")` first.

Do you want me to add calls to `chdir` (or `SetDir`) before calls to `fopen` that don't have them? Primarily this would be in the `audio/sampman_*.cpp` source files.

erorcun: oh sorry for late reply.
I would prefer moving what you're doing right now on CFileMgr::OpenFile to ChangeDir/SetDir, since doing them in there is kinda hack.

Tatsh: My `OpenFile` change has been removed. I realised it was only there to fix an issue with `ValidateVersion()` in `main.cpp`.

I made `ValidateVersion()` call `SetDir("")` before doing anything else if XDG_ROOT is enabled. Can this be done without the `#ifdef`?

Tested the game from a save file and from new game and everything seems normal.

src/core/FileMgr.cpp
```diff
@@ -197,10 +197,69 @@ myfeof(int fd)
...
+// Must have char homeDir[PATH_MAX] in scope.
+#define getenvvar(varName)                                  \
+	char **p = environ;                                     \
+	size_t varNameLength = ARRAY_SIZE(varName) - 1;         \
+	for (; *p; p++) {                                       \
+		if (!strncmp(varName "=", *p, varNameLength + 1)) { \
+			strcpy(homeDir, *p + varNameLength + 1);        \
+			break;                                          \
+		}                                                   \
+	}
```
Tatsh, line 210: Not sure why these \ do not line up on GitHub. They line up in my IDE:

erorcun:
I remind you your message:
> Correct, there is no chdir call to set the path to the XDG path. So, it is possible for an fopen call to fail if for example, cSampleManager::InitialiseSampleBanks() (opens audio/sfx.*) was called without a call to CFileMgr::SetDir("") / CFileMgr::SetDir("\\") first.

but looks like you haven't done that. We *have* to call chdir on ChangeDir/SetDir. Game may work ok right now but this doesn't mean that it's not coincidence.

Tatsh:
> I remind you your message:
> 
> > Correct, there is no chdir call to set the path to the XDG path. So, it is possible for an fopen call to fail if for example, cSampleManager::InitialiseSampleBanks() (opens audio/sfx.*) was called without a call to CFileMgr::SetDir("") / CFileMgr::SetDir("\") first.
> 
> but looks like you haven't done that. We _have_ to call chdir on ChangeDir/SetDir. Game may work ok right now but this doesn't mean that it's not coincidence.

I think these should solve the issue:

https://github.com/GTAmodding/re3/pull/746/commits/3bd9e61a12724171d4b521a0ac4718c2d7d40095
https://github.com/GTAmodding/re3/pull/746/commits/1e245d755b927748b80f3d4fd2aed2a5246533c2
https://github.com/GTAmodding/re3/pull/746/commits/273e4e8a3b11c2b20b73c49f0cbaecd82f81a155

kattjevfel:
Unsure if this is out of scope for this PR, but would this allow for setting any directory? For example it'd be great to be able to install to /usr/share/re3 with the binary in /usr/bin.

Tatsh:
That wouldn't be for here but it's the same idea. I think we could have a system version or a more configurable build to allow setting the path dynamically.

I am doing this path because it's part of an open standard and should not require root privileges to write to.

src/core/FileMgr.cpp
sjerdo: Tab indent vs spaces

PR #751, opened 2020-10-08
Add utility to install GTA III game files

Tatsh:
~~WIP and Linux/BSD only for now.~~

Not for Windows.

The CLI installer will run like so:

```
installer-command disc1.iso disc2.iso
```

or with mounted discs or ISOs:

```
installer-command path-to-disc-1/ path-to-disc-2/
```

This uses 7z to extract ISOs if necessary, and then uses libunshield to unpack the `data1.cab` file on disc 1. Then `cp` is invoked directly to copy the `Audio` directory from disc 2.

This can work on Windows but there will need some changes (like not invoking `cp` and use of `fork`), and I may complete that. However, I am thinking on Windows a better solution is to create an NSIS-based installer application to do this task.

@erorcun Please examine `.clang-format` and see if this really matches the project's standards.


erorcun:
We decided to not include that in re3, sorry. But it would be cool to see it as a seperate tool.

PR #761, opened 2020-10-11
SampleManager

erorcun:
Now with usertrack support in Linux :)


src/audio/AudioManager.cpp
```diff
@@ -618,29 +618,56 @@ cAudioManager::AddDetailsToRequestedOrderList(uint8 sample)
...
 
+// --MIAMI: Done
 void
 cAudioManager::AddReflectionsToRequestedQueue()
 {
+#ifdef FIX_BUGS
+  	uint32 oldFreq = 0;
+#else
+  	uint32 oldFreq;
+#endif
```
ShFil119, line 629: iirc msvc won't compile uninitialized code like this (with disabled `FIX_BUGS`)

ShFil119 (approved):

src/audio/AudioManager.cpp
erorcun: just tried it, it does

ShFil119: my bad

PR #763, opened 2020-10-11
SpecialFX

theR4K:


Fire-Head:
you missed this fix in CTimer::Update

`		int32 updInCycles = (pc.LowPart - _oldPerfCounter.LowPart); // & 0x7FFFFFFF; pointless
		
		// Don't scale when in-menu
		float updInCyclesScaled = GetIsPaused() ? updInCycles : updInCycles * ms_fTimeScale;
		
		_oldPerfCounter = pc;`

src/render/SpecialFX.cpp
```diff
@@ -516,7 +762,7 @@ C3dMarkers::Render()
 					CCoronas::RegisterCorona((uintptr)&m_aMarkerArray[i],
 						SPHERE_MARKER_R, SPHERE_MARKER_G, SPHERE_MARKER_B, 192,
 						m_aMarkerArray[i].m_Matrix.GetPosition(), 1.2f * m_aMarkerArray[i].m_fSize, 50.0f * TheCamera.LODDistMultiplier,
-						CCoronas::TYPE_STAR, CCoronas::FLARE_NONE, CCoronas::REFLECTION_OFF, CCoronas::LOSCHECK_OFF, CCoronas::STREAK_ON, 0.0f, false);
+						CCoronas::TYPE_STAR, CCoronas::FLARE_NONE, CCoronas::REFLECTION_OFF, CCoronas::LOSCHECK_OFF, CCoronas::STREAK_ON, 0.0f, false, 1.5f);
```
aap, line 765: not needed, it's default

src/render/SpecialFX.cpp
```diff
@@ -788,6 +1027,13 @@ CBrightLights::Render(void)
 			g = aBrightLights[i].m_green;
 			b = aBrightLights[i].m_blue;
 			break;
+		default:
+#ifdef FIX_BUGS  //just to make sure that color never will be undefined
```
aap, line 1031: if you absolutely want to handle this case, just return.

src/render/SpecialFX.cpp
```diff
@@ -869,8 +1131,8 @@ CBrightLights::Render(void)
 
 		case BRIGHTLIGHT_SIREN:
 			for(j = 0; j < 6; j++){
-				pos = SirenLightsSide[j]*aBrightLights[i].m_side +
-					SirenLightsUp[j]*aBrightLights[i].m_up +
+				pos = SirenLightsSide[j]*TheCamera.GetRight() +
```
aap, line 1134: fix this in re3 as well, even though it's unused in both games.

src/render/SpecialFX.cpp
```diff
@@ -851,6 +1097,22 @@ CBrightLights::Render(void)
 			TempBufferIndicesStored += 12*3;
 			break;
 
+		case BRIGHTLIGHT_FRONT_BIG:
+		case BRIGHTLIGHT_REAR_BIG:
+			for (j = 0; j < 8; j++) {
```
aap, line 1102: fix this in re3 too. how tf did i miss this?

src/render/SpecialFX.cpp
```diff
@@ -1066,20 +1329,15 @@ CMoneyMessage::Render()
...
 			CFont::SetPropOn();
 			CFont::SetBackgroundOff();
-
-			float fScaleY = fDistY / 100.0f;
-			if (fScaleY > MAX_SCALE) fScaleY = MAX_SCALE;
-
-			float fScaleX = fDistX / 100.0f;
-			if (fScaleX > MAX_SCALE) fScaleX = MAX_SCALE;
-
+			float fScaleY = Min(fDistY / 100.0f, MAX_SCALE);
```
aap, line 1332: in re3 too

src/render/SpecialFX.h
```diff
@@ -37,33 +45,31 @@ class CMotionBlurStreaks
...
 struct CBulletTrace
 {
 	CVector m_vecCurrentPos;
 	CVector m_vecTargetPos;
 	bool m_bInUse;
-	uint8 m_framesInUse;
-	uint8 m_lifeTime;
+	uint32 m_nCreationTime;
+	uint32 m_nLifeTime;
+	float m_fThicknes;
```
aap, line 56: as above, this is m_fSpeed

src/render/SpecialFX.cpp
```diff
@@ -229,10 +327,12 @@ CMotionBlurStreaks::RegisterStreak(uintptr id, uint8 r, uint8 g, uint8 b, CVecto
 			return;
 		}
 	}
+
 	// Find free slot
-	for(i = 0; aStreaks[i].m_id; i++)
+	for(i = 0; aStreaks[i].m_id != nil ; i++)
```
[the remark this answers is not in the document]
theR4K: but it has a pointer type!
if you want I can change to !=0. But let this without compare it's bad idea because it's not a boolean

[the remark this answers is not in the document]
aap: no, it's uintptr, which is an integer type. if anything use != 0 but i don't see why that would be necessary, you'd only have to change re3 as well.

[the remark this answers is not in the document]
theR4K: ok then will be better if I change it in re3 too. Bcs it's types and their meaning

erorcun (approved):

PR #775, opened 2020-10-18
Added logo

hazelnot:


erorcun:
can you replace it with the red one, and move them to root folder? we're not thinking to use it as an app icon right now, but we can put it to Readme etc.

hazelnot:
> 
> 
> can you replace it with the red one, and move them to root folder? we're not thinking to use it as an app icon right now, but we can put it to Readme etc.

That was actually the original version I made, based on the European logo, but I figured aesthetically the one based on the remaster is more aesthetically pleasing. I can replace it with the red version though.

PR #778, opened 2020-10-20
Fluff done, FireInstantHit fix 

majesticCoding:


src/render/Fluff.cpp
```diff
@@ -391,66 +384,47 @@ int32 CScrollBar::TonightsEvent;
...
+	for (int i = 0; i < NUMSECTORS_X; i++) {
+		for (int j = 0; j < NUMSECTORS_Y; j++) {
+			for (CPtrNode *pNode = CWorld::GetSector(i, j)->m_lists[ENTITYLIST_BUILDINGS].first; pNode; pNode = pNode->next) {
+				CEntity *pEntity = (CEntity *)pNode->item;
+				PossiblyAddThisEntity(pEntity);
+			}
+		}
+	}
+
+	for (int32 i = 0; i < 3; i++) {
```
erorcun, line 413: let's add NUM_LEVELS to eLevelName and replace 3 with it

src/render/Fluff.cpp
```diff
@@ -504,41 +467,84 @@ void CMovingThings::Render()
...
-			aTowerClocks[i].Render();
-	}
-	for (i = 0; i < ARRAY_SIZE(aDigitalClocks); ++i)
-	{
-		if (aDigitalClocks[i].IsVisible())
-			aDigitalClocks[i].Render();
+
+	for (i = 0; i < Num; i++) {
+		if (aMovingThings[i].m_nHidden == 0)
+			aMovingThings[i].Update();
```
erorcun, line 473: this shouldn't be in this func.

src/render/Fluff.cpp
```diff
@@ -504,41 +467,84 @@ void CMovingThings::Render()
...
 }
 
 // ---------- CMovingThing ----------
+float lengths[5] = { 100.0f, 1500.0f, 400.0f, 100.0f, 2000.0f };
+
 void CMovingThing::Update()
 {
+	switch (m_nType) {
+	case 1: {
+		float angle = (CTimer::GetTimeInMilliseconds() % 0x3FFF) * 0.0003835f;
```
erorcun, line 488: 0.0003835f = TWOPI / 0x3FFF

src/render/Fluff.cpp
```diff
@@ -504,41 +467,84 @@ void CMovingThings::Render()
...
+
+		if (CClock::GetHours() >= 20 || CClock::GetHours() < 5) {
+			if (Abs(TheCamera.GetPosition().x - m_pEntity->GetPosition().x) < 600.0f &&
+				Abs(TheCamera.GetPosition().y - m_pEntity->GetPosition().y) < 600.0f) {
+				CVector delta = m_pEntity->GetPosition() - TheCamera.GetPosition();
+				delta.Normalise();
+
+				if (delta.x * Cos(angle) + delta.y * Sin(angle) < -0.92f) {
+					CVector coors = m_pEntity->GetPosition() - 10.0f * delta;
+					CCoronas::RegisterCorona((uintptr)&m_pEntity + 1, 128, 128, 100, 255, coors, 70.0f, 600.0f, 0.0f, CCoronas::TYPE_STAR, CCoronas::REFLECTION_OFF, CCoronas::LOSCHECK_OFF, CCoronas::STREAK_OFF, 0.0f, false, 1.5f);
```
erorcun, line 501: first parameter should be just 43

src/render/Fluff.cpp
```diff
@@ -504,41 +467,84 @@ void CMovingThings::Render()
...
+				}
+			}
+		}
+	}
+		break;
+	case 2: {
+		float angle = (CTimer::GetTimeInMilliseconds() % 0x7FF) * 0.003068f;
```
erorcun, line 508: this is TWOPI / 0x7FF

src/render/Fluff.cpp
```diff
@@ -504,41 +467,84 @@ void CMovingThings::Render()
...
+		m_pEntity->GetRight() = CVector(Cos(angle), Sin(angle), 0.0f);
+		m_pEntity->GetForward() = CVector(-Sin(angle), Cos(angle), 0.0f);
+		m_pEntity->GetUp() = CVector(0.0f, 0.0f, 1.0f);
+	}
+		break;
+	case 3: {
+		float angle = (CTimer::GetTimeInMilliseconds() % 0x3FF) * 0.006136f;
```
erorcun, line 515: TWOPI / 0x3FF

src/render/Fluff.cpp
```diff
@@ -504,41 +467,84 @@ void CMovingThings::Render()
...
+		m_pEntity->GetRight() = CVector(Cos(angle), Sin(angle), 0.0f);
+		m_pEntity->GetForward() = CVector(-Sin(angle), Cos(angle), 0.0f);
+		m_pEntity->GetUp() = CVector(0.0f, 0.0f, 1.0f);
+	}
+		break;
+	case 4: {
+		float angle = (CTimer::GetTimeInMilliseconds() % 0x3FFFF) * 0.000024f;
```
erorcun, line 522: TWOPI / 0x3FFFF

src/render/Fluff.cpp
```diff
@@ -568,27 +574,31 @@ int16 CMovingThing::SizeList()
...
 
-// ---------- Find message functions ----------
-const char* FindTunnelMessage()
-{
-	if (CStats::CommercialPassed)
-		return "LIBERTY TUNNEL HAS BEEN OPENED TO ALL TRAFFIC . . .   ";
-
-	if (CStats::IndustrialPassed)
-		return "FIRST PHASE LIBERTY TUNNEL HAS BEEN COMPLETED . . .   ";
+void CMovingThings::RegisterOne(int16 nType, CEntity *pEnt) {
```
erorcun, line 577: pls use original signature

src/render/Fluff.cpp
```diff
@@ -568,27 +574,31 @@ int16 CMovingThing::SizeList()
...
-const char* FindBridgeMessage()
-{
-	if (CStats::CommercialPassed)
-		return "STAUNTON LIFT BRIDGE IS OPERATIONAL AGAIN    ";
+void CMovingThings::PossiblyAddThisEntity(CEntity *pEnt) {
+	if (pEnt->GetModelIndex() == MI_LIGHTBEAM)
+		RegisterOne(1, pEnt);
+
+	if (pEnt->GetModelIndex() == MI_AIRPORTRADAR)
+		RegisterOne(2, pEnt);
```
erorcun, line 594: also those `if`s after the first if should be `else if`

src/render/Fluff.cpp
```diff
@@ -601,34 +611,8 @@ const char* FindTimeMessage()
...
-		String_DigitalClock[2] = ' ';
-		String_DigitalClock[3] = '@';
-		String_DigitalClock[4] = 'C';
-	}
-	return String_DigitalClock;
-}
-
 // ---------- CScrollBar ----------
-void CScrollBar::Init(CVector position, uint8 type, float sizeX, float sizeY, float sizeZ, uint8 red, uint8 green, uint8 blue, float scale)
+void CScrollBar::Init(CVector pos1, CVector pos2, uint8 type, uint8 red, uint8 green, uint8 blue, float scale)
```
erorcun, line 615: looks like there is some missing bits in this func. and after doing it you will notice struct has changed since III

src/render/Fluff.cpp
```diff
@@ -672,263 +656,46 @@ void CScrollBar::Update()
...
-						"ALSO CRAIG GRAYS) DEBUT, )FADE AWAY), INCLUDES THE SINGLE OF THE SAME NAME. . . ";
-					break;
-				case 8:
-					m_pMessage =
-						"ON THE FILM FRONT, A NELY COMPILED COMPILATION OF ARNOLD STEELONES GREATEST MOVIES ON DVD. "
-						"THE PACK INCLUDES THE EARLY )BY-CEP), THE CULT CLASSIC )FUTURE ANNHILATOR), AND THE HILARIOUS CROSS-DRESSING COMEDY )SISTERS). "
-						"ONE FOR ALL THE FAMILY. . . ";
-					break;
-				case 9:
+				default:
```
erorcun, line 694: this is case 3. default is just `break`.

erorcun (changes requested):

src/render/Fluff.cpp
majesticCoding: can't see any other fields or smth like that
The only distinction I see is that the size assignment, which is scaled difference between two points now

majesticCoding: the string doesn't work correct with it
looks like it takes strlen from ". ", not from m_pMessage, hence it equals 2. The same thing in gta3

src/render/Fluff.cpp
```diff
@@ -568,27 +569,32 @@ int16 CMovingThing::SizeList()
...
 
-// ---------- Find message functions ----------
-const char* FindTunnelMessage()
-{
-	if (CStats::CommercialPassed)
-		return "LIBERTY TUNNEL HAS BEEN OPENED TO ALL TRAFFIC . . .   ";
-
-	if (CStats::IndustrialPassed)
-		return "FIRST PHASE LIBERTY TUNNEL HAS BEEN COMPLETED . . .   ";
+void CMovingThings::RegisterOne(CEntity *pEnt, uint8 nType) {
```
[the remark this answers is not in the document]
majesticCoding: fixed

src/render/Fluff.cpp
erorcun: ? isn't m_pMessage already ". "? I mean it's obvious that it will always be 2, but you should do it in original way

majesticCoding: yeah, fixed

erorcun (approved):

PR #780, opened 2020-10-23
ObjectData

Sergeanur:


src/objects/ObjectData.cpp
```diff
@@ -18,11 +20,55 @@ CObjectData::Initialise(const char *filename)
...
+	ms_aObjectInfo[0].m_fTurnMass = 99999.0f;
+	ms_aObjectInfo[0].m_fAirResistance = 0.99f;
+	ms_aObjectInfo[0].m_fElasticity = 0.1f;
+	ms_aObjectInfo[0].m_fBuoyancy = GRAVITY * ms_aObjectInfo[0].m_fMass * 2.0f;
+	ms_aObjectInfo[0].m_fUprootLimit = 0.0f;
+	ms_aObjectInfo[0].m_fCollisionDamageMultiplier = 1.0f;
+	ms_aObjectInfo[0].m_nCollisionDamageEffect = 0;
+	ms_aObjectInfo[0].m_nSpecialCollisionResponseCases = 0;
+	ms_aObjectInfo[0].m_bCameraToAvoidThisObject = false;
+
```
Bigbossbro08, line 34: I think it's better if it's wrapped with a function/constructor

Sergeanur: Don't think original did that.

Bigbossbro08: Point but I guess it can make code look little better? or was it inlined?

aap (approved):

src/objects/ObjectData.cpp
Sergeanur: I don't think it's inlined. We don't stray too much unless there a reason.

erorcun (approved):

ShFil119 (approved):

PR #792, opened 2020-10-29
fix FINAL build, add DEBUGMENU and other simple qol defines to FINAL ...

withmorten:
… add extra defines for menu and game version text.


src/core/timebars.cpp
```diff
@@ -1,5 +1,5 @@
 #include "common.h"
-#ifndef MASTER
+#ifndef FINAL
```
Sergeanur, line 2: Don't think so since the code of functions is on mobile

withmorten: Forced push away, was an accidentally leftover from MASTER testing.

Sergeanur (approved):

PR #797, opened 2020-11-01
Make code in master same like in audio PR for miami

theR4K:


aap:
All changes were better before.

theR4K:
> All changes were better before.

I'm not agree, for first: 
default it's special case when another cases not entered and it's very important to place it after all others cases(unless you want make code unreadable, then it's ok!). In other words, lets allow compiler to to his job!
second one: using event variable it's good for understanding AND for fault tolerance, because if something change event in array, when this function is processing. It

theR4K:
c

PR #801, opened 2020-11-03
new LCS-style renderer

aap:
a bit drafty, but i wanna get it out


aap:
pushed this directly now.

PR #803, opened 2020-11-05
add bInvertLook4Pad from VC

withmorten:
just some preliminary addition of this possibilty. What would be the best way to save this? break .set compability once again? :D


erorcun:
No. we have CFO for that. You can do it in a one line.

withmorten:
I updated it to do it via CFO, and added a custom string for it in american.gxt :)

src/core/MenuScreensCustom.cpp
```diff
@@ -780,6 +787,9 @@ CMenuScreenCustom aScreens[MENUPAGES] = {
    { "FET_MTI", MENUPAGE_CONTROLLER_PC, MENUPAGE_CONTROLLER_PC, nil, nil,
 	   MENUACTION_MOUSESENS,	"FEC_MSH",	{ nil, SAVESLOT_NONE, MENUPAGE_MOUSE_CONTROLS },
 	   MENUACTION_INVVERT,		"FEC_IVV",	{ nil, SAVESLOT_NONE, MENUPAGE_MOUSE_CONTROLS },
+#ifdef INVERT_LOOK_FOR_PAD
```
erorcun, line 790: you can remove that ifdef, you already wrapped INVERT_PAD_SELECTOR with it

withmorten: Right, I just saw that you did it everywhere that way too, so I wanted to keep your coding style. Should I keep or remove it?

PR #810, opened 2020-11-08
error on strat app

hamed34434:
hi
this version (07.11.2020 Patch)  not working error after strat app on atmosphere 0.15.0
atmosphere 0.14.4 worked


aap:
Are you PRing the switch port? this makes no sense.

PR #815, opened 2020-11-13
CPed review part 1

aap:
Some things were bugs, lots of flipped conditions, some other things closer to original code, some stylistic changes.

### DIFF
--- a/src/peds/Ped.cpp
+++ b/src/peds/Ped.cpp
@@ -140,9 +140,6 @@ bool CPed::bFannyMagnetCheat;
 bool CPed::bPedCheat3;
 CVector2D CPed::ms_vec2DFleePosition;
 
-CVector vecNextPathNode;
-bool vecNextPathNodeInitialized;
-
 void *CPed::operator new(size_t sz) { return CPools::GetPedPool()->New();  }
 void *CPed::operator new(size_t sz, int handle) { return CPools::GetPedPool()->New(handle); }
 void CPed::operator delete(void *p, size_t sz) { CPools::GetPedPool()->Delete((CPed*)p); }
@@ -287,8 +284,7 @@ CPed::CPed(uint32 pedType) : m_pedIK(this)
 	m_phoneId = -1;
 	m_lastAccident = 0;
 	m_fleeFrom = nil;
-	m_fleeFromPosX = 0;
-	m_fleeFromPosY = 0;
+	m_fleeFromPos = CVector2D(0.0f, 0.0f);
 	m_fleeTimer = 0;
 	m_threatEx = nil;
 	m_vecSeekPosEx = CVector(0.0f, 0.0f, 0.0f);
@@ -722,24 +718,11 @@ CPed::CanStrafeOrMouseControl(void)
 void
 CPed::AddWeaponModel(int id)
 {
-	RpAtomic *atm;
-
 	if (id != -1) {
-#ifdef PED_SKIN
-		if (IsClumpSkinned(GetClump())) {
-			if (m_pWeaponModel)
-				RemoveWeaponModel(-1);
+		if (m_pWeaponModel)
+			RemoveWeaponModel(-1);
 
-			m_pWeaponModel = (RpAtomic*)CModelInfo::GetModelInfo(id)->CreateInstance();
-		}
-		else
-#endif
-		{
-			atm = (RpAtomic*)CModelInfo::GetModelInfo(id)->CreateInstance();
-			RwFrameDestroy(RpAtomicGetFrame(atm));
-			RpAtomicSetFrame(atm, m_pFrames[PED_HANDR]->frame);
-			RpClumpAddAtomic(GetClump(), atm);
-		}
+		m_pWeaponModel = (RpAtomic*)CModelInfo::GetModelInfo(id)->CreateInstance();
 		CModelInfo::GetModelInfo(id)->AddRef();
 		m_wepModelID = id;
 
@@ -752,7 +735,6 @@ CPed::AddWeaponModel(int id)
 void
 CPed::AimGun(void)
 {
-	RwV3d pos;
 	CVector vector;
 
 	if (IsPlayer() && bIsDucking)
@@ -760,8 +742,7 @@ CPed::AimGun(void)
 
 	if (m_pSeekTarget) {
 		if (m_pSeekTarget->IsPed()) {
-			((CPed*)m_pSeekTarget)->m_pedIK.GetComponentPosition(pos, PED_MID);
-			vector = pos;
+			((CPed*)m_pSeekTarget)->m_pedIK.GetComponentPosition(vector, PED_MID);
 		} else {
 			vector = m_pSeekTarget->GetPosition();
 		}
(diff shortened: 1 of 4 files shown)


Sergeanur (approved):

src/peds/Ped.cpp
```diff
@@ -19642,21 +19544,20 @@ CPed::WarpPedToNearLeaderOffScreen(void)
...
-		if (Abs(zCorrectedPos.z - warpToPos.z) >= 3.0f && Abs(zCorrectedPos.z - appropriatePos.z) >= 3.0f)
-			continue;
-
-		appropriatePos.z = zCorrectedPos.z;
-		if (!TheCamera.IsSphereVisible(appropriatePos, 0.6f, &TheCamera.GetCameraMatrix())
-			&& CWorld::GetIsLineOfSightClear(appropriatePos, warpToPos, true, true, false, true, false, false, false)
-			&& !CWorld::TestSphereAgainstWorld(appropriatePos, 0.6f, this, true, true, false, true, false, false)) {
-			teleported = true;
-			Teleport(appropriatePos);
+		if (Abs(zCorrectedPos.z - warpToPos.z) < 3.0f || Abs(zCorrectedPos.z - appropriatePos.z) < 3.0f) {
```
erorcun, line 19553: this is a matter of style, not a review

src/peds/Ped.cpp
```diff
@@ -19607,21 +19510,20 @@ CPed::WarpPedToNearEntityOffScreen(CEntity *warpTo)
...
 	CVector appropriatePos = GetPosition();
-	CVector zCorrectedPos = appropriatePos;
-	int tryCount = Min(10, halfOfDist);
+	int tryCount = Min(10, (int)halfOfDist);
 	for (int i = 0; i < tryCount; ++i) {
 		appropriatePos += halfNormalizedDist;
+		CVector zCorrectedPos = appropriatePos;
 		CPedPlacement::FindZCoorForPed(&zCorrectedPos);
 
-		if (Abs(zCorrectedPos.z - warpToPos.z) >= 3.0f && Abs(zCorrectedPos.z - appropriatePos.z) >= 3.0f)
```
erorcun, line 19616: this is my style, not a mistake

src/peds/Ped.cpp
```diff
@@ -10714,7 +10653,7 @@ CPed::MoveHeadToLook(void)
 
 	if (m_pLookTarget) {
 		if (m_pLookTarget->IsPed()) {
-			((CPed*)m_pLookTarget)->m_pedIK.GetComponentPosition(*(RwV3d *)&lookPos, PED_MID);
```
erorcun, line 10717: this is here because either Linux or Win was creating problems

src/peds/Ped.cpp
```diff
@@ -9317,18 +9264,15 @@ CPed::FollowPath(void)
 CVector
 CPed::GetFormationPosition(void)
 {
-	CPed *referencePed = m_pedInObjective;
```
erorcun, line 9320: this is my style...

src/peds/Ped.cpp
```diff
@@ -7120,9 +7072,7 @@ CPed::CreateDeadPedMoney(void)
 	if (!CGame::nastyGame)
 		return;
 
-	int skin = GetModelIndex();
-
-	if ((skin >= MI_COP && skin <= MI_FIREMAN) || (CharCreatedBy == MISSION_CHAR && !bMoneyHasBeenGivenByScript) || bInVehicle)
+	if ((GetModelIndex() >= MI_COP && GetModelIndex() <= MI_FIREMAN) || (CharCreatedBy == MISSION_CHAR && !bMoneyHasBeenGivenByScript) || bInVehicle)
```
erorcun, line 7075: this is your style either

src/peds/Ped.cpp
```diff
@@ -2802,11 +2773,8 @@ CPed::SetPedStats(ePedStats pedStat)
 bool
 CPed::CanUseTorsoWhenLooking(void)
 {
-	if (m_nPedState != PED_DRIVING && m_nPedState != PED_DRAG_FROM_CAR && !bIsDucking) {
-		if (m_animGroup != ASSOCGRP_SEXYWOMAN && m_animGroup != ASSOCGRP_WOMAN)
-			return true;
-	}
-	return false;
```
erorcun, line 2809: again this is not a review...

src/peds/Ped.cpp
```diff
@@ -9579,11 +9518,11 @@ CPed::GoToNearestDoor(CVehicle *veh)
 bool
 CPed::HaveReachedNextPointOnRoute(float distToCountReached)
 {
-	if ((m_nextRoutePointPos - GetPosition()).Magnitude2D() >= distToCountReached)
-		return false;
```
erorcun, line 9583: wow much review

src/peds/Ped.cpp
aap: comparing floats for equality only makes sense in specific cases, this is not one of them. we can be 100% sure the original code had < here and hence this is simply original R* code.

src/peds/Ped.cpp
aap: then at least we need a better name, `skin` makes no sense.

src/peds/Ped.cpp
aap: it makes the code harder to understand. and you weren't even being consistent: `return m_pedInObjective->GetMatrix() * formationOffset;`

src/peds/Ped.cpp
aap: see comment on floats above

src/peds/Ped.cpp
aap: ok, this one i don't insist one.

src/peds/Ped.cpp
erorcun: 100% sure? no, it only makes sense to you and people aren't machines that obey to rules. i just dug up my old codes and I've found date comparisons with >= and <= for example, i LIKE using <= and >= and it's possible to use them

src/peds/Ped.cpp
erorcun: you're right on the latter one, but it just makes it harder to YOU to understand! in how many ways should I express myself?

src/peds/Ped.cpp
erorcun: ok

src/peds/Ped.cpp
aap: dates aren't floats. most floating point values are inexact and == makes little sense. when you have something that you happen to know is expressible as the sum of powers of two (like integers), then comparing with == makes sense (like some == 99999.0f checks we have in various places), but not here. and R* didn't write it like this. I didn't not change this because i didn't like your style, in fact i might have written it same as you, but we want to stick closely to R* code.

src/peds/Ped.cpp
aap: i still don't see what's the point of referencePed though. and the assignment `m_pedInObjective = referencePed  = ...` is particularly awkward.

src/peds/Ped.cpp
erorcun: but no one can know that. as an example to floats, here is some map code on my pending Frontend PR:
```
if(m_fMapSize >= MENU_Y(1000.0f) && in) \
			break; \
```

I wrote that >= willingly, and consciously. why shouldn't i use this? what harm can this do?

PR #821, opened 2020-11-16
CPed review part2

aap:
like before...

### DIFF
--- a/src/peds/Ped.cpp
+++ b/src/peds/Ped.cpp
@@ -287,8 +287,8 @@ CPed::CPed(uint32 pedType) : m_pedIK(this)
 	m_fleeFromPos = CVector2D(0.0f, 0.0f);
 	m_fleeTimer = 0;
 	m_threatEx = nil;
-	m_vecSeekPosEx = CVector(0.0f, 0.0f, 0.0f);
-	m_distanceToCountSeekDoneEx = 0.0f;
+	m_vecSpotToGuard = CVector(0.0f, 0.0f, 0.0f);
+	m_radiusToGuard = 0.0f;
 	m_nWaitState = WAITSTATE_FALSE;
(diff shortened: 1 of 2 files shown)


src/peds/Ped.cpp
```diff
@@ -13909,8 +13898,7 @@ CPed::PedSetQuickDraggedOutCarPositionCB(CAnimBlendAssociation *animAssoc, void
 	if (ped->m_vehEnterType == CAR_DOOR_RF || ped->m_vehEnterType == CAR_DOOR_RR)
 		draggedOutOffset.x = -draggedOutOffset.x;
 
-	finalLocalPos = Multiply3x3(pedMat, draggedOutOffset);
-	finalPos = finalLocalPos + ped->GetPosition();
+	finalPos = Multiply3x3(pedMat, draggedOutOffset) + ped->GetPosition();
```
erorcun, line 13901: is this original? if so can you show me how can i understand it? except that lgtm

aap: CVector functions use lots of temporaries. a1 below here surely is one so we only have one original variable here
```
      Multiply3x3(&a1, &v18, &a3a);
      finalPos = a1;
      finalPos.x = a1.x + a3->_._._._.m_matrix.m_matrix.pos.x;
      finalPos.y = a1.y + a3->_._._._.m_matrix.m_matrix.pos.y;
      finalPos.z = a1.z + a3->_._._._.m_matrix.m_matrix.pos.z;
```

erorcun: output variables of all Multiply3x3 calls are temporary then?

aap: Hm, probably, yes.

erorcun: ok

PR #824, opened 2020-11-19
Made a snap folder for building snap packages

0x0000ff:
Hello there,
Thank you for getting GTAIII to run on newer platforms.
I have made a snap package builder for you. It has a desktop launcher that can run the game when stored in 
- ~/snap/re3/common.
- /var/snap/re3/common
I have the second option because I have a root partition on a SSD and /home on a HDD

\- 0x0000ff


0x0000ff:
Sorry, I neglected to mention a few things. There is currently a mouse cursor in the center of the screen when in game, but I'm pretty sure it can be removed easily as other snap packaged games do not have this problem.

0x0000ff:
I neglected to mention that there's a mouse pointer visible in the center of the screen when in game. I'm sure this is easily amended in the snapcraft.yaml file, as this isn't an issue on other snap packaged games.

You'll most likely have to change file permissions for the game folder as well if it's stored in /var/snap/re3/common.

ShFil119:
How can I test it? It'd be best to add worker to github actions for it?

kusst (changes requested):

0x0000ff:
`snapcraft try` and `snap try --devmode` in the root of the repository. You'll need to install both of the programs through snap, and it will use a small VM based on 20.04 to build it with (but you can use an LXD container instead).

The snap builder shouldn't effect the rest of the project when building it and vice versa. However the snap won't build if you change a library or add another one without listing it in the snapcraft.yaml

PR #833, opened 2020-11-23
red warning texts in delete save menu

withmorten:
I just deleted my main save on accident, because I accidentally entered this menu and went into muscle memory mode. These red warning texts should throw you off enough to NOT delete your save while you think you're loading it.

The red doesn't look perfect, but then of course it doesn't need to, it just needs to get your attention :)

re3 2020-11-24 00-24-48-68

### DIFF
--- a/src/core/Frontend.cpp
+++ b/src/core/Frontend.cpp
@@ -1049,6 +1049,18 @@ CMenuManager::Draw()
 	if (aScreens[m_nCurrScreen].m_ScreenName[0] != '\0') {
 		
 		PREPARE_MENU_HEADER
+#ifdef RED_DELETE_TEXTS
+		switch(m_nCurrScreen){
+		case MENUPAGE_CHOOSE_DELETE_SLOT:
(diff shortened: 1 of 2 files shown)


Calinou:
Maybe the menu itself could be changed to make it harder to accidentally delete a save file?

For instance, you could be required to enter a string into a text field like "confirm".

withmorten:
@erorcun you had the idea to maybe draw a red background too/instead? Honestly I think this is good enough, nobdy ever looks at this menu on purpose anyway.

erorcun:
> @erorcun you had the idea to maybe draw a red background too/instead? Honestly I think this is good enough, nobdy ever looks at this menu on purpose anyway.

yeah, I don't have time to do it anyway. I still think it doesn't look good, but i'm ok if others are ok with it. i'll wait for them

withmorten:
Apparently this won't get merged anytime soon, or at all, so I'll close this until a better idea comes around :P

PR #838, opened 2020-11-25
implemented CMemoryHeap

aap:


src/core/MemoryHeap.cpp
```diff
@@ -0,0 +1,484 @@
...
+		else {
+			LoadingScreen("NO MORE MEMORY", nil, nil);
+			LoadingScreen("NO MORE MEMORY", nil, nil);
+		}
+		CGame::TidyUpMemory(true, false);
+	}
+	void *mem = Malloc(size);
+	if (removeCollision) {
+		CTimer::Stop();
+		// different on PS2
```
Fire-Head, line 183: `#ifdef GTA_PS2
CFileLoader::LoadCollisionFile((CCollision::ms_pColFiles[CCollision::ms_collisionInMemory]);
#else`
maybe then ?

src/rw/MemoryHeap.cpp
```diff
@@ -0,0 +1,484 @@
...
+	memcpy(dst, ptr, block->m_size);
+	Free(ptr);
+	return dst;
+}
+
+void
+CMemoryHeap::Free(void *ptr)
+{
+	HeapBlockDesc *block = GetDescFromHeapPointer(ptr);
+	MEMORYHEAP_ASSERT(block->m_memId != MEMID_FREE);
```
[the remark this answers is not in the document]
aap: will fix in next commit after merging

src/core/MemoryHeap.cpp
aap: it's not really ps2 specific though, just older code. and we haven't found a very good way to kept track of older code yet. besides, it would need changes in CCollision as well to be consistent. so i think i'll leave it for now until we have decided what to do about these cases.

PR #844, opened 2020-11-28
Case sensitivity fix

ZLau92:
Case sensitivity fix.
Fixes compile issue on linux after recent changes with splitting up script.cpp


Sergeanur:
Merged, thanks

Issue #782, opened 2020-10-24
Miami: multiple issues for passing storyline


ga2mer:
You may be aware of these problems, but as I progress through the game, I found several problems:
- Crash on `Rub Out` mission in cutscene when Diaz looked at cameras (useless getting vehicle position I think)
<details>
  <summary>Rub Out backtrace</summary>

  ```
Thread 1 "reVC" received signal SIGSEGV, Segmentation fault.
0x000055555573f878 in CPed::PositionAnyPedOutOfCollision (this=0x555557366850) at ../src/peds/Ped.cpp:14229
14229           CVector vehPos = m_pMyVehicle->GetPosition();
(gdb) bt
#0  0x000055555573f878 in CPed::PositionAnyPedOutOfCollision (this=0x555557366850) at ../src/peds/Ped.cpp:14229
#1  0x00005555557d987c in CRunningScript::ProcessCommands100To199 (this=0x555555cc4050 <CTheScripts::ScriptsArray+11088>, command=161) at ../src/control/Script.cpp:3777
#2  0x00005555557d3a05 in CRunningScript::ProcessOneCommand (this=0x555555cc4050 <CTheScripts::ScriptsArray+11088>) at ../src/control/Script.cpp:2518
#3  0x00005555557d32da in CRunningScript::Process (this=0x555555cc4050 <CTheScripts::ScriptsArray+11088>) at ../src/control/Script.cpp:2456
#4  0x00005555557d314b in CTheScripts::Process () at ../src/control/Script.cpp:2423
#5  0x00005555556a55a8 in CGame::Process () at ../src/core/Game.cpp:751
#6  0x0000555555899a0e in Idle (arg=0x1) at ../src/core/main.cpp:1023
#7  0x000055555589a113 in AppEventHandler (event=rsIDLE, param=0x1) at ../src/core/main.cpp:1276
#8  0x00005555558a4426 in RsEventHandler (event=rsIDLE, param=0x1) at ../src/skel/skeleton.cpp:212
#9  0x0000555555896696 in main (argc=1, argv=0x7fffffffdd68) at ../src/skel/glfw/glfw.cpp:1870
  ```
  
</details>

- ~Windows not count as shattered on `Shakedown` mission where you need broke multiple windows in North Point Mall~
- ~After window shattering, collision still exists and this stops passing `G-Spotlight` mission, where you need to jump from one building to other through window
[...]

PR #847, opened 2020-11-29
CObject fixes

waliedyassen:
Fixes 1 bug with `CObject::DeleteAllTempObjectsInArea`, implements 3 missing behaviors for `MI_RCBOMB`, `MI_BEACHBALL`  and `MI_PETROLPUMP2` and all the missing damage effects.

This should fix `The Shootist` mission problem in  #782.


src/core/General.h
```diff
@@ -159,6 +159,11 @@ class CGeneral
 		
 	static int32 GetRandomNumberInRange(int32 low, int32 high)
 		{ return low + (high - low)*(GetRandomNumber()/float(MYRAND_MAX + 1)); }
+
+	// Returns inclusive value in the specified range
+	static int32 GetRandomNumberInRangeInc(int32 low, int32 high)
+		{ return GetRandomNumberInRange(low - 1, high + 1); }
+
```
Sergeanur, line 166: We don't need this.

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
 		m_vecMoveSpeed *= fTimeStep;
 		m_vecTurnSpeed *= fTimeStep;
 	}
-	if ((GetModelIndex() == MI_EXPLODINGBARREL || GetModelIndex() == MI_PETROLPUMP) && bHasBeenDamaged && bIsVisible
+	auto mi = GetModelIndex();
```
Sergeanur, line 112: We don't use auto.

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+	if (mi == MI_RCBOMB) {
+		float fTurnForce = -(m_fTurnMass / 20.0f);
+		CPhysical::ApplyTurnForce(m_vecMoveSpeed * fTurnForce, -GetForward());
+		float fScalar = 1.0f - m_vecMoveSpeed.MagnitudeSqr() / 5.0f;
+		float fScalarTimed = Pow(fScalar, CTimer::GetTimeStep());
+		m_vecMoveSpeed *= fScalarTimed;
+	}
+	if (mi == MI_BEACHBALL) {
+		constexpr uint8 BEACHBALL_MAX_SCORE = 250;
+		constexpr float BEACHBALL_DEACCELERATION = 2.5f;
```
Sergeanur, line 130: We don't use constexpr. Use #define instead.

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+		float fMoveSpeedMag = m_vecMoveSpeed.Magnitude2D();
+		float fTimeScale = powf(0.95, CTimer::GetTimeStep());
```
Sergeanur, line 132: `Pow`. Also in here and other occasions below floating point value needs `f` suffix

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+			float distanceMagnitude = distance.Magnitude2D();
+			if (distance.z > 0.0 && distance.z < 1.5f && distanceMagnitude < 1.0) {
+				CVector playerSpeed = FindPlayerSpeed();
+				if (fMoveSpeedMag < 0.05 && playerSpeed.Magnitude() > 0.1) {
+					playerSpeed.z = 0.0f;
+					playerSpeed.Normalise();
+					playerSpeed.z = 0.3;
+					m_vecMoveSpeed = CVector(playerSpeed.x / BEACHBALL_DEACCELERATION, playerSpeed.y / BEACHBALL_DEACCELERATION, 1.0f / BEACHBALL_DEACCELERATION * 0.3);
+					PlayOneShotScriptObject(SCRIPT_SOUND_HIT_BALL, GetPosition());
+					if (m_nBeachballBounces > 0) {
```
Sergeanur, line 147: Missing turn speed calculation above this

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+						m_nBeachballBounces++;
+						sprintf(gString, "%d", m_nBeachballBounces);
```
Sergeanur, line 149: Missing another `if (m_nBeachballBounces > 0) {` above this.

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+		constexpr uint8 BEACHBALL_MAX_SCORE = 250;
+		constexpr float BEACHBALL_DEACCELERATION = 2.5f;
+		float fMoveSpeedMag = m_vecMoveSpeed.Magnitude2D();
+		float fTimeScale = powf(0.95, CTimer::GetTimeStep());
+		m_vecMoveSpeed.x *= fTimeScale;
+		m_vecMoveSpeed.y *= fTimeScale;
+		m_vecMoveSpeed.z += fMoveSpeedMag - m_vecMoveSpeed.Magnitude2D();
+		if (!FindPlayerVehicle()) {
+			CVector distance = FindPlayerCoors() - GetPosition();
+			float distanceMagnitude = distance.Magnitude2D();
```
Sergeanur, line 138: A bit of a nit, but original code didn't use variable in here and instead called `distance.Magnitude2D()` both times.

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+						m_nBeachballBounces++;
+						sprintf(gString, "%d", m_nBeachballBounces);
+						CMoneyMessages::RegisterOne(GetPosition(), gString, 255, 50, 0, 0.6f, 0.5f);
+						CStats::RegisterHighestScore(3, m_nBeachballBounces);
+					}
+				}
+			}
+			if (distance.z > -1.05 && distance.z < -0.6 && distanceMagnitude < 0.9 && m_vecMoveSpeed.z < 0.0f) {
+				m_vecMoveSpeed.x += CGeneral::GetRandomNumberInRangeInc(-3, 4) / 100.0;
+				m_vecMoveSpeed.y += CGeneral::GetRandomNumberInRangeInc(-3, 4) / 100.0;
```
Sergeanur, line 157:
```suggestion
				m_vecMoveSpeed.x += (CGeneral::GetRandomNumber() % 8 - 3) / 100.0f;
				m_vecMoveSpeed.y += (CGeneral::GetRandomNumber() % 8 - 3) / 100.0f;
```

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+				m_vecMoveSpeed.z = Max(m_vecMoveSpeed.z + 0.3f, 0.2f);
+				PlayOneShotScriptObject(SCRIPT_SOUND_HIT_BALL, GetPosition());
+				m_vecTurnSpeed.x += CGeneral::GetRandomNumberInRangeInc(-7, 8) / 10.0f;
+				m_vecTurnSpeed.y += CGeneral::GetRandomNumberInRangeInc(-7, 8) / 10.0f;
```
Sergeanur, line 161:
```suggestion
				m_vecTurnSpeed.x += (CGeneral::GetRandomNumber() % 16 - 7) / 10.0f;
				m_vecTurnSpeed.y += (CGeneral::GetRandomNumber() % 16 - 7) / 10.0f;
```

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+				if (++m_nBeachballBounces >= BEACHBALL_MAX_SCORE) {
+					m_nBeachballBounces = BEACHBALL_MAX_SCORE;
+				}
```
Sergeanur, line 164:
```suggestion
				m_nBeachballBounces++;
				m_nBeachballBounces = Max(m_nBeachballBounces, BEACHBALL_MAX_SCORE);
```

src/objects/Object.cpp
```diff
@@ -107,14 +109,65 @@ CObject::ProcessControl(void)
...
+				sprintf(gString, "%d", m_nBeachballBounces);
+				CMoneyMessages::RegisterOne(GetPosition(), gString, 255, 50, 0, 0.6f, 0.5f);
+				CStats::RegisterHighestScore(3, m_nBeachballBounces);
+			}
+		}
+	}
```
Sergeanur, line 170: Some flag code is missing in the end of function.

Sergeanur (changes requested):
Thank you for your contribution, there's a couple of things to be fixed though.

src/objects/Object.cpp
ShFil119: or just `static const`

Sergeanur: But not in this occasion, please.

src/objects/Object.cpp
waliedyassen: Max wouldn't work there, I assume you meant Min instead

Sergeanur: Indeed

src/objects/Object.cpp
```diff
@@ -287,132 +349,320 @@ CObject::ObjectDamage(float amount)
...
+					CParticle::AddParticle(PARTICLE_CAR_DEBRIS, vecPos, vecDir, nil, fSize, selectedColor, nRotationSpeed, 0, nCurFrame, 0);
+				}
+				if (m_nCollisionDamageEffect == DAMAGE_EFFECT_SMASH_BLACKBAG) {
+					PlayOneShotScriptObject(SCRIPT_SOUND_BOX_DESTROYED_2, vecPos);
+				} else if (m_nCollisionDamageEffect == DAMAGE_EFFECT_SMASH_BEACHLOUNGE_WOOD) {
+					PlayOneShotScriptObject(SCRIPT_SOUND_METAL_COLLISION, vecPos);
+				}
+				break;
+			}
+			default:
```
aap, line 663: where is DAMAGE_EFFECT_SMASH_VEGPALM?

src/objects/Object.cpp
```diff
@@ -107,19 +112,76 @@ CObject::ProcessControl(void)
...
+	if (mi == MI_BEACHBALL) {
+		float fTimeStep = Pow(0.95f, CTimer::GetTimeStep());
+		float fPreviousVecSpeedMag = m_vecMoveSpeed.Magnitude2D();
+		m_vecMoveSpeed.x *= fTimeStep;
+		m_vecMoveSpeed.y *= fTimeStep;
+		m_vecMoveSpeed.z += fPreviousVecSpeedMag - m_vecMoveSpeed.Magnitude2D();
+		if (!FindPlayerVehicle()) {
+			CVector distance = FindPlayerCoors() - GetPosition();
+			float distanceMagnitude = distance.Magnitude2D();
+			if (distance.z > 0.0 && distance.z < 1.5f && distanceMagnitude < 1.0) {
```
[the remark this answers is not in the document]
waliedyassen: It's used in the other condition, in the mobile binary they have it as a variable like this.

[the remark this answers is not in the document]
aap: That's a compiler optimization. the android compiler was much more aggressive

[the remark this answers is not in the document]
waliedyassen: Makes sense

src/objects/Object.cpp
```diff
@@ -287,132 +349,320 @@ CObject::ObjectDamage(float amount)
...
+			case DAMAGE_EFFECT_SMASH_COMPLETELY:
+			case DAMAGE_EFFECT_CHANGE_THEN_SMASH:
+			{
+				if (m_nCollisionDamageEffect == DAMAGE_EFFECT_CHANGE_THEN_SMASH && !bRenderDamaged) {
+					bRenderDamaged = true;
+					return;
+				}
 				bIsVisible = false;
 				bUsesCollision = false;
+				if (!GetIsStatic()) {
```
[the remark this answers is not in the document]
waliedyassen: Are you sure about that one? Looks like it is the way it is right now for me.

src/objects/Object.cpp
```diff
@@ -287,132 +349,320 @@ CObject::ObjectDamage(float amount)
...
+			case DAMAGE_EFFECT_SPLIT_MODEL:
+				return;
+			case DAMAGE_EFFECT_SMASH_AND_DAMAGE_TRAFFICLIGHTS:
+			{
+				static RwRGBA debrisColor = { 0xc8,0xc8,0xc8,0xff };
+				if (bRenderDamaged) {
+					break;
+				}
 				bRenderDamaged = true;
+				CBaseModelInfo *modelInfo = CModelInfo::GetModelInfo(GetModelIndex());
```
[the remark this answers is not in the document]
waliedyassen: Will do, but that would not match the binary anymore, to be more accurate, in the binary it's something like this:
```cpp
CVector min = 0.85f * CModelInfo::GetModelInfo(GetModelIndex())->GetColModel()->boundingBox.min;
CVector max = 0.85f * CModelInfo::GetModelInfo(GetModelIndex())->GetColModel()->boundingBox.max;
```

src/objects/Object.cpp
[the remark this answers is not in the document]
aap: oh yeah, i got confused. logically this makes no sense to me. why would you remove non-static objects from the moving list???

src/objects/Object.cpp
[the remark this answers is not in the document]
aap: you don't have to go through modelinfo manually, GetColModel takes care of that. but yes, it looks like they didn't even use a temporary for the col model, so just
```
CVector min = 0.85f * GetColModel()->boundingBox.min;
CVector max = 0.85f * GetColModel()->boundingBox.max;
`

File: src/objects/Object.cpp [hunk 16]
```diff
@@ -14,15 +14,20 @@
...
 
 int16 CObject::nNoTempObjects;
 //int16 CObject::nBodyCastHealth = 1000;
 float CObject::fDistToNearestTree;
 
-void *CObject::operator new(size_t sz) { return CPools::GetObjectPool()->New();  }
-void *CObject::operator new(size_t sz, int handle) { return CPools::GetObjectPool()->New(handle);};
-void CObject::operator delete(void *p, size_t sz) { CPools::GetObjectPool()->Delete((CObject*)p); }
-void CObject::operator delete(void *p, int handle) { CPools::GetObjectPool()->Delete((CObject*)p); }
+void *CObject::operator new(size_t sz) { return CPools::GetObjectPool()->New(); }
```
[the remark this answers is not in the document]
waliedyassen: This was a revert to the style changes I did in a previous commit, I'll go through and make sure it is matching.

src/core/General.h
```diff
@@ -159,6 +159,7 @@ class CGeneral
 		
 	static int32 GetRandomNumberInRange(int32 low, int32 high)
 		{ return low + (high - low)*(GetRandomNumber()/float(MYRAND_MAX + 1)); }
+
```
Sergeanur, line 162: Please remove this empty line change.

waliedyassen: Done

Sergeanur: Eh, now there's a change at the end of it (

waliedyassen: That matches the miami branch, the line was an extra line as well

Sergeanur: It wouldn't show it as a change if it's a match.

waliedyassen: Sorry, you are right. fixed it. But that's weird, it only shows when checking out locally. The empty lines are removed when viewing in the GitHub code viewer ([like here](https://github.com/GTAmodding/re3/blob/miami/src/core/General.h))

Sergeanur (approved):

Issue #845, opened 2020-11-28
Translation additions


IlDucci:
Greetings, I would like to ask a few questions before attempting to do some PRs for you.

First of all, I would like to know if you would accept translation corrections for the existing strings or just translations for the new RE3 ones. The Spanish translation (which is the one I would like to work with) of GTA3 is notoriously bad and it's been kept like that ever since (the iOS port just cleaned up the biggest offenders, but without checking the large amount of mistranslations that exist with the English version).

Secondly, I would like to ask if you are planning on adding extra strings for pads/controllers support. The PC version doesn't have a controller menu with the images of the PS2/whatever controller (I could try to help with the graphical part as well), but the thing I'd like to point out is that all the help messages were written thinking on PC; and thus, any action that's handled by moving a stick appears on screen as "Press the X button to do Y." Are you planning on adding extra strings to retrieve the PS2's stick-based messages and port them into PC?

Sergeanur:
Wouldn't hurt if you fix any translation issues you could find. You shouldn't worry about stick-as-a-button issue. The strings that mention sticks are kept as is on PC and only used in HELP4_D and HELP5_D, they show up when you play with control setup 4. Btw, PC version actually has the controller menu that shows DS2 (see #454), it's hidden and a bit unfinished, though it would be cool to have a variation of it for Xbox controller and DS4.

PR #849, opened 2020-11-29
Spanish translation: adding new re3 strings.

IlDucci:
This is a minor update to separate the new strings from re3 from the actual Spanish translation revamp that I've mentioned in https://github.com/GTAmodding/re3/issues/845 .

Contains the re3 strings, the iOS strings as well as fixes one stupid error.


Sergeanur:
The iOS strings are redundant in here, we're not using them. Only those that are below `{ re3 updates }` should be added.

IlDucci:
I've removed the iOS strings.

PR #850, opened 2020-11-29
GTA_VERSION define and some config.h cleanup

aap:
i hope there are no mistakes...


src/core/config.h
```diff
@@ -173,14 +180,18 @@ enum Config {
...
 
+// Memory allocation and compression
+// #define USE_CUSTOM_ALLOCATOR		// use CMemoryHeap for allocation. use with care, not finished yet
+//#define COMPRESSED_COL_VECTORS	// use compressed vectors for collision vertices
+//#define ANIM_COMPRESSION	// only keep most recently used anims uncompressed
+
 #if defined GTA_PS2
 #	define GTA_PS2_STUFF
 #	define RANDOMSPLASH
+#	define USE_CUSTOM_ALLOCATOR
```
Sergeanur, line 191: Do we need this?

aap: PS2 does use CMemoryHeap. maybe you can do without it on a TOOL, but that seems more like an exception.

Sergeanur: Ah, I see

src/core/config.h
```diff
@@ -65,6 +65,7 @@ enum Config {
 	NUMCULLZONES = 512,
 	NUMATTRIBZONES = 288,
 	NUMZONEINDICES = 55000,
+//	NUMZONEINDICES = 140000*3,
```
erorcun, line 68: debug leftover I think

src/audio/sampman_miles.cpp
```diff
@@ -1297,17 +1297,17 @@ cSampleManager::Terminate(void)
 bool
 cSampleManager::CheckForAnAudioFileOnCD(void)
 {
-#if !defined(GTA3_STEAM_PATCH) && !defined(NO_CDCHECK)
+#if GTA_VERSION < GTA3_PC_STEAM && !defined(NO_CDCHECK)
 	char filepath[MAX_PATH];
 	
-#if defined(GTA3_1_1_PATCH)
+#if GTA_VERSION >= GTA3_PC_STEAM
```
Sergeanur, line 1303: Shouldn't this be `GTA_VERSION >= GTA3_PC_11`?

Issue #843, opened 2020-11-28
Unable to control camera pitch while in vehicle /w Gamepad


ZLau92:
On Pop OS 20.10 x64

While in-vehicle with free cam enabled, I was only able to move the camera left/right, but not up/down with the right-stick of my gamepad.

I narrowed it down to this bit of code:
https://github.com/GTAmodding/re3/blob/f50a53e2907bcadd5c89781c4d9c025a45c8a113/src/core/Cam.cpp#L5035-L5037
With those lines commented out, the freecam works as intended with a gamepad.

The comment in the referenced bit of code reads:
> // In SA this checks for m_bUseMouse3rdPerson so num2/num8 do not move camera when Keyboard & Mouse Controls are used.

But I wasn't able to induce any undesirable effects, neither by gamepad nor key/mouse with these lines commented.

ZLau92:
Fixed
https://github.com/GTAmodding/re3/commit/33e1f28239e715aba0abbf417bbcf57e61985d65

Issue #685, opened 2020-08-12
Add joypad support for the debug menu


AGraber:
Debug menu is currently uncontrollable on setups with joypad support only (consoles, etc.) so a way for navigating the debug menu with joypad would be great.

I tried to take a peek but the code is a bit confusing, doing lots of hardcoded checks everywhere, probably was made without this in mind.

PR #855, opened 2020-11-30
Gamepad fixes, SDL2 Config & Initial debug menu support

ZLau92:
Sorry for the duplicate pull request, just getting used to version controlling.
#685 
The highlight here is I've gotten some initial gamepad compatibility for the debug menu. More details about that in the commit.

Also a pull request from earlier with SDL gamepad configs from environment variables.

Plus a couple cam/anim bugfixes I've come across.
#843


erorcun:
You should mark/wrap changes that's done to original code, this is a reverse engineering oriented project.
so I'm not in favor of adding `\n` to Renderer.cpp. also you should explain your change in Automobile.cpp. other changes look ok to me, but I think @aap should decide it since he is the author of debugmenu.

ZLau92:
Ok, duly noted.

ZLau92:
@erorcun thanks, and apologies for any confusion as I become familiar with the process of contributing in general.
For Automobile.cpp, the change I made was to address an issue whereby the driveby_r animation wouldn't play unless you have 0 passengers or are in first person mode. I removed the if() portion of the else if clause that was causing this to happen.
I wasn't able to replicate this behavior in the ps2 versions of iii/vc.

ZLau92:
Looks like this PR may have gotten a bit messy. I'm going to close it and break these changes out.

Issue #839, opened 2020-11-25
Ultrawide monitor with "Auto" aspect ratio will heavily reduce traffic


consolethinks:
I have a 3440x1440 monitor, and when the game is set to this resolution with "Auto" aspect ratio, the game will display everything correctly, but the traffic spawn rate will get heavily reduced. This does not seem to be affected by the frame limiter. Both vanilla and miami branches are affected by this.

Since there's no aspect ratio setting in the original game (aside from Widescreen ON/OFF), I can't test it there.

Calinou:
> Since there's no aspect ratio setting in the original game (aside from Widescreen ON/OFF), I can't test it there.

Can you reproduce this in the vanilla game with the [widescreen fix](http://thirteenag.github.io/wfp#gta) isntalled?

consolethinks:
> Can you reproduce this in the vanilla game with the widescreen fix installed?
I tried the widescreen fix with both GTA III and Vice City, using forced 21:9 aspect ration with the same resolution, and the bug doesn't get reproduced. 

In re3 (vanilla/miami), this bug is most apparent with traffic coming from the front, but it affects spawning from all directions, just maybe less so with off screen spawns.

ZLau92:
I don't have an ultra-wide, but I do have a dual 1440p monitor setup. I set to windowed mode and dragged the window to fill both monitors, then set the aspect ratio to auto, and the lack of traffic was quite apparent within the first few minutes of driving around.
The FOV is so wide at a certain point, the game engine may think that you have almost 360 degrees of observed state around you.

ZLau92:
> he game engine may think that you have almost 360 degrees of observed state around you.
I was totally wrong. It turns out this happens because of the generation distance value, which uses the aspect ratio in it's calculation... And it gets a little wonky if you go much wider than 16:9, trying to place cars further away than the max spawn distance.

ZLau92:
Fixed a long time ago, can be closed
https://github.com/GTAmodding/re3/commit/b8bc54640d80db3d0c0a6392a0858d83a09e8304

PR #856, opened 2020-12-01
Update Camera.cpp - Vehicle spawning on ultrawide aspect ratios

ZLau92:
#839 
camdist multiplier is consitently between 1 and 1.4 on 16:9 aspect ratio
On ultrawide, the cam dist multiplier jumps to anywhere between 2 and 3
The preferred distance generator for car objects is the cam dist multiplier * 120. Which puts cars that it's trying to generate outside of the allowed gen/draw distance for new car objects when on ultrawide. This means no new cars can spawn.
This change caps the scaling for the generation distance a bit above what it would be for 16:9 and does not affect the LOD multiplier.

### DIFF
--- a/src/core/Camera.cpp
+++ b/src/core/Camera.cpp
@@ -720,7 +720,7 @@ CCamera::Process(void)
 	else
 		LODDistMultiplier = 1.0f;
 	// missing on PS2
-	GenerationDistMultiplier = LODDistMultiplier;
+	GenerationDistMultiplier = 70.0f/CDraw::GetFOV() * fmin(CDraw::GetAspectRatio(),1.82f)/(4.0f/3.0f);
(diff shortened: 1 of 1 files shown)


ZLau92:
Question: Does the comment above the line of code I modified ("missing on PS2") indicate that it's not original code?
Happy to wrap my changes, but want to make sure it's necessary. It might also be worth noting that the original bug report indicated that this issue does not occur on the vanilla version of the game, even with a forced high aspect ratio, so I'd be willing to bet the original (or at least ps2) version just uses a dist multiplier of 1.

aap:
There are no LOD dist multipliers on PS2 (at least in III) is all it says. now that we have  GTA_VERSION define it'll be easier to wrap all the differences between PS2 and PC in defines, even though it gets a bit messy :/

PR #860, opened 2020-12-02
Another fix proposal for ultrawide

ZLau92:
@aap pushed a fix this morning, but it still seems to have the same trouble with cars not spawning.
It seems that this was happening for a couple of reasons:
1. This fix sets the LOD Distance to 2.2 and adds bug fix / ps2 version wraps, but it seems that the values are overwriting eachother depending on the config (ie, if you are both above the PS2 release & with fix_bugs)
2. The GenerationDistMultiplier really shouldn't be > ~ 1.36, and the code here appears to cap it at 2.2 per the loddist multiplier. I'm not wholly convinced these values need to be tied together in this context past that val.

So with both of these things in minds, I've modified the wraps to have an if/elif that should be kosher to the original code, and I've set a cap for the gendist var to 1.36 under fix_bugs, while still allowing the higher 2.2f max loddist mltplr from vice city.

### DIFF
--- a/src/core/Camera.cpp
+++ b/src/core/Camera.cpp
@@ -720,7 +720,10 @@ CCamera::Process(void)
 #ifdef FIX_BUGS
         // from VC. to high values bug out spawns
         LODDistMultiplier = Min(LODDistMultiplier, 2.2f);
-        GenerationDistMultiplier = Min(LODDistMultiplier, 1.36f);
+        #if GTA_VERSION > GTA3_PS2_160
+        // The LODDistMultiplier bugfix can apply to PS2, but the GenerationDistMultiplier isn't used in that build
+                GenerationDistMultiplier = Min(LODDistMultiplier, 1.36f);
+        #endif
 #elif GTA_VERSION > GTA3_PS2_160
         GenerationDistMultiplier = LODDistMultiplier;
         LODDistMultiplier *= CRenderer::ms_lodDistScale;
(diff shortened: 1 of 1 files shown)


aap:
no, this is wrong. with your PR we don't do `LODDistMultiplier *= CRenderer::ms_lodDistScale;` when FIX_BUGS is on.
i assumed the 2.2 would just work in III, but apparently it doesn't. What i think we should do is just force GenerationDistMultiplier to 1.0 and be done with it. Why should it change with FOV and aspect ratio anyway?

aap:
Ok i think we fixed it for good now. The aspect ratio scale is actually not there in VC so i removed it under FIX_BUGS in III as well. and there was another problem in pathfind. it hope it's working well for you now.

ZLau92:
Hey there, @aap
I think you may have misread my PR. `LODDistMultiplier *= CRenderer::ms_lodDistScale;` doesn't occur under BUG_FIX wrap, but under the version wrap as it did in your PR.

I'll have a quick look later, but it seems that your fix is still scaling the gendist with the aspect ratio/fov, and still has the same spawn problem. Perhaps the gendist should just be capped at 1.0
Btw, you can test for this issue by setting to windowed mode and pulling it down to half-height, then driving down a straight stretch of road.

ZLau92:
Nvm! I'm wrong. You fixed it.
I forgot to pull your change after resetting mine.

PR #862, opened 2020-12-02
Major editing and proofread of the official Spanish translation 

IlDucci:
Alright, this is the big mother update I mentioned. Ref: https://github.com/GTAmodding/re3/issues/845

This commit includes a ton of changes and fixes to the original Spanish translation. Some have been taken from the iOS version (which was just a cleanup), others were done by myself.

I've reorganized the entire file per missions and sections just so I could know where I was looking at.

There's too much changes to list them, so here's the short version:
 - Retranslated and naturalized a lot of overly literal translations.
 - Fixed some uppercasing and lowercasing issues in the menus.
 - Shortened some subtitle translations that appear in a very short span of time so they are readable by the player.
 - This translation is European Spanish based, just like the original one and following Rockstar's policy up until GTAV. However, I tried to "neutralize" characters that are Latino American (El Burro, Catalina, Miguel, the Colombians...). If someone wants to do a fully Neutral Spanish translation, it should use a new file instead of overwriting this one in a translation war.


Sergeanur:
The string reshuffle is unwelcomed, they should be in the original order.

IlDucci:
Damn, that's a pity. Okay, I'll undo the reordering... Is there any technical reasoning behind this sorting? Just being curious.

Sergeanur:
The changes could be easily maintained that way.

utils/gxt/spanish.txt
```diff
@@ -7919,25 +7896,25 @@ INVERTIR VERTICALIDAD RATÓN
...
-CONTROLES: ESTÁNDAR
+CONTROL: ESTÁNDAR
 
 [FES_SET]
-Usar skin
+USAR APARIENCIA
 
 [GHOST]
-Ghost
+Fantasma
```
Sergeanur, line 7908: Does a boat name has to be translated?

IlDucci:
> 
> 
> The changes could be easily maintained that way.

If that's the only reason and there's no more (the entire file is kept in memory, and the "sorting" is automatically done in the header pieces, not the text pieces), then I would suggest that all the assets would be resorted in a human-understandable way, like I did. I won't do that because the easiest way to do this would be with some scripting, but I'm no programmer. I did the sorting and unsorting by hand.

utils/gxt/spanish.txt
IlDucci: > 
> 
> Does a boat name has to be translated?

Nope. I thought it was something else, so I didn't recognise this as a vehicle name.

cirno-999: fun fact: Stallion (car) is actually translated in the Polish TL (as Ogier.)
I think it was translated separately from the main development though, so there may be a few quirks (it's still p damn good.)

Sergeanur:
I'd like to have them in R*'s order, and ideally transfer comments from leaked iOS source too. The strings changed at late development stage were moved to the bottom.

PR #870, opened 2020-12-06
Fixes for Linux

madebr:
Hello,

While working on the cmake build script + a conan script,
I had to apply these patches to get it building on Linux.

- @ShFil119 
  This fixes your issue at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97728
  [`inttypes.h` is part of the c standard](https://en.cppreference.com/w/cpp/header/cinttypes)
- The function `casepath` is required on non-Windows systems, so include `crossplatform.h`.
  Its use does not depend on the support of opus.


ShFil119:
Thanks

PR #885, opened 2020-12-15
Few bug fixes in CAutomobile

waliedyassen:


aap:
looks good, but while you're at it, can you remove the now stupid comment above the CARWHEEL enum?

waliedyassen:
> looks good, but while you're at it, can you remove the now stupid comment above the CARWHEEL enum?

Removed

PR #886, opened 2020-12-15
Radio scroll to prev. station & radio off text

erorcun:
Just "off" text actually


src/audio/MusicManager.cpp
```diff
@@ -83,12 +83,41 @@ cMusicManager::DisplayRadioStationName()
...
+		if(gNumRetunePresses < 0) {
+			gStreamedSound = m_nCurrentStreamedSound;
+
+			if(gStreamedSound == STREAMED_SOUND_CITY_AMBIENT ||
+			   gStreamedSound == STREAMED_SOUND_WATER_AMBIENT) {
+				gStreamedSound = POLICE_RADIO; // which means OFF
+			} else {
+
+				if(gStreamedSound >
+				   STREAMED_SOUND_RADIO_MP3_PLAYER)
```
Nick007J, line 96: Is newline here intended? seems a bit odd here

Nick007J (approved):

src/audio/MusicManager.cpp
erorcun: I've copied it from the code below (shfil's?). changed both of them now

ShFil119 (approved):

PR #891, opened 2020-12-19
Attempt to merge master to miami

Nick007J:
Please check if it looks adequate, it compiles so far, I'll check more defines later


src/audio/sampman_miles.cpp
```diff
@@ -1380,7 +1380,7 @@ cSampleManager::Terminate(void)
 bool
 cSampleManager::CheckForAnAudioFileOnCD(void)
 {
-#if !defined(GTA3_STEAM_PATCH) && !defined(NO_CDCHECK)
+#if !defined(NO_CDCHECK) // TODO: figure out defines
```
erorcun, line 1383: why?

src/core/AnimViewer.cpp
```diff
@@ -47,7 +47,7 @@ CEntity *CAnimViewer::pTarget = nil;
 void
 CAnimViewer::Render(void) {
 	if (pTarget) {
-//		pTarget->GetPosition() = CVector(0.0f, 0.0f, 0.0f);
+//		pTarget->GetPosition() = CVector(0.0f, 0.0f, 0.0f); // Only on Mobile
```
erorcun, line 50: needs confirmation if this is also correct for VC mobile

src/core/Cam.cpp
```diff
@@ -5053,9 +5053,9 @@ CCam::Process_FollowCar_SA(const CVector& CameraTarget, float TargetOrientation,
 	float stickX = -(pad->GetCarGunLeftRight());
 	float stickY = pad->GetCarGunUpDown();
 
-	// In SA this checks for m_bUseMouse3rdPerson so num2/num8 do not move camera when Keyboard & Mouse controls are used.
-	if (CCamera::m_bUseMouse3rdPerson)
-		stickY = 0.0f;
+	// In SA this is for not let num2/num8 move camera when Keyboard & Mouse controls are used.
+	// if (CCamera::m_bUseMouse3rdPerson)
+	//	stickY = 0.0f;
```
erorcun, line 5058: these shouldn't be commented for miami, iirc VC uses num 2/8 unlike III

src/core/config.h
```diff
@@ -361,9 +363,15 @@ static_assert(false, "SUPPORT_XBOX_SCRIPT and SUPPORT_MOBILE_SCRIPT are mutually
...
 
 //#define SQUEEZE_PERFORMANCE
 #ifdef SQUEEZE_PERFORMANCE
 	#undef PS2_ALPHA_TEST
 	#undef NO_ISLAND_LOADING
 #endif
+
+#ifdef LIBRW
+// these are not supported with librw yet
+#endif
```
erorcun, line 377: we already have this above

src/peds/Ped.h
```diff
@@ -1104,6 +1103,16 @@ class CPed : public CPhysical
...
+	inline void SetWeaponLockOnTarget(CEntity *target)
+	{
+		if (m_pPointGunAt)
+			m_pPointGunAt->CleanUpOldReference(&m_pPointGunAt);
+
+		m_pPointGunAt = (CPed*)target;
+		if (target)
+			((CEntity*)target)->RegisterReference(&m_pPointGunAt);
+	}
+
```
erorcun, line 1115: this is not inlined on VC. Serge assumed it should've been inlined in III, so added it to header for master

src/render/Renderer.cpp
```diff
@@ -24,11 +24,16 @@
...
+#include "Debug.h"
 
 //--MIAMI: file done
 
+bool gbShowPedRoadGroups;
+bool gbShowCarRoadGroups;
 bool gbShowCollisionPolys;
 bool gbShowCollisionLines;
+bool gbShowCullZoneDebugStuff;
+bool gbDisableZoneCull;	// not original
```
erorcun, line 36: cullzone in miami?

src/peds/Ped.h
Sergeanur: It's actually inlined everywhere except mobile, most likely due to touchscreen controls.

Nick007J: It has to be inlined since it is for CCopPed constructor at least

src/audio/sampman_miles.cpp
Nick007J: Yeah I'll add a keyword, probably it would be GTAVC_STEAM_PATCH but needs confirmation

src/control/Record.cpp
```diff
@@ -69,12 +69,17 @@ void CRecordDataForChase::ProcessControlCars(void)
 {
 }
 
-#if (defined(GTA_PS2) || defined(FIX_BUGS))
 bool CRecordDataForChase::ShouldThisPadBeLeftAlone(uint8 pad)
 {
+	// may be wrong
```
aap, line 74: all of the record functions are empty in VC. i simply deleted all the code, maybe ifdefs would have been better, but having this one function implemented is awkward.

src/core/Game.cpp
```diff
@@ -354,21 +390,35 @@ bool CGame::Initialise(const char* datFile)
...
+	POP_MEMID();
+
 	InitModelIndices();
+
+	PUSH_MEMID(MEMID_DEF_MODELS);
 	CModelInfo::Initialise();
 	CPickups::Init();
 	CTheCarGenerators::Init();
+
+#ifndef GTA_PS2		// or GTA_VERSION?
```
aap, line 418: checked VC now. it's GTA_VERSION in III, VC always has it.

src/core/Pools.cpp
```diff
@@ -27,19 +28,38 @@ CDummyPool *CPools::ms_pDummyPool;
 CAudioScriptObjectPool *CPools::ms_pAudioScriptObjectPool;
 CColModelPool *CPools::ms_pColModelPool;
 
+#ifdef GTA_PS2	// or USE_CUSTOM_ALLOCATOR
```
aap, line 31: I don't see this in VC, but it could have been moved into MASTER or something

src/modelinfo/SimpleModelInfo.h
```diff
@@ -11,6 +11,18 @@ class CSimpleModelInfo : public CBaseModelInfo
 	float  m_lodDistances[3];
 	uint8  m_numAtomics;
 	uint8  m_alpha;
+	/* // For reference, PS2 has:
```
aap, line 14: Not sure if this is still the case in VC actually....

src/render/Renderer.cpp
```diff
@@ -901,6 +911,14 @@ CRenderer::RenderCollisionLines(void)
 	}
 }
 
+// unused
+void
+CRenderer::RenderBlockBuildingLines(void)
```
aap, line 916: not to be found in VC

src/peds/Ped.h
erorcun: so you guys think they moved it out of header in VC mobile?

Nick007J: To my knowledge, compiler technically cannot inline a function to another object file unless it has a body in header (unless link-time optimization is on). On the other hand, it may ignore inline keyword and presense of function in a header. So I don't think it's possible for this function not to be in header. Whether it was moved to cpp on mobile, or just compiler decided it's better to call it, we can't know for certain.

erorcun: hm, but how one of the SetObjectives ended up inlined on our build and we had to force it to not to be inlined then?

erorcun: oh you said another object file. ok lgtm

PR #894, opened 2020-12-21
signed/unsigned fixes, and some other fixes

erorcun:


src/core/Fire.cpp
```diff
@@ -399,13 +398,12 @@ CFireManager::ExtinguishPointWithWater(CVector point, float range)
 	for (int i = 0; i < NUM_FIRES; i++) {
 		if (m_aFires[i].m_bIsOngoing) {
 			if ((point - m_aFires[i].m_vecPos).MagnitudeSqr() < sq(range)) {
-				fireI = i;
 				break;
 			}
 		}
+		if (++fireI >= NUM_FIRES)
```
Nick007J, line 404: Very strange code... Though I feel like `i < NUM_FIRES` wasn't there originally. On PS2 and mobile it looks like a while loop with the `!bIsOngoing || range-check`, which doesn't look right either.

Nick007J (approved):

PR #907, opened 2020-12-29
Fixing an original issue: removing unneeded white color codes

IlDucci:
This is an original issue. It seems that certain cutscenes had a white color control code in their subtitles when
 - It's not needed for subtitles.
 - The `~w~` control code isn't pure white, but light gray, so those subtitles look different than the rest.

I've also fixed a couple of unwanted spaces found in French and German, and also tweaked one string in Spanish and added the translation for the new Waypoint string.


IlDucci:
Changes unwelcomed by the core team.

Calinou:
> Changes unwelcomed by the core team.

Out of curiosity, why is that?

IlDucci:
The team considered those ~w~ codes to be placed on purpose and not leftovers that should be removed. I'm going to do another PR to fix actual typos.

Calinou:
Superseded by https://github.com/GTAmodding/re3/pull/923.

PR #909, opened 2020-12-29
Sprite Scaling Option

Fire-Head:


src/render/Draw.h
```diff
@@ -5,9 +5,21 @@ enum eAspectRatio
 	// Make sure these work the same as FrontEndMenuManager.m_PrefsUseWideScreen
 	// without widescreen support
 	AR_4_3,
+	AR_5_4,
+	AR_16_10,
 	AR_16_9,
+	AR_21_9,
 
 	AR_AUTO,
```
Sergeanur, line 13: As this enum change breaks settings compatibility anyway we should move AR_AUTO to the top of it.

utils/gxt/american.txt
```diff
@@ -8058,15 +8058,24 @@ NEO
...
 PS2
 
+[FEM_PC]
+PC
+
 [FEM_XBX]
 XBOX
 
+[FEM_AUT]
+AUTO
```
Sergeanur, line 8068: Worth using it with aspect ratio option too

erorcun:
can you merge Hud Scaling and PS2_HUD options? I mean having 2 options are confusing, and PC scaling + PS2_HUD is *not* PS2_HUD, am I wrong?

Fire-Head:
i would like to have PS2 scaling(+radar/coronas fix) for PC hud too, since this is the only correct combination, but not everyone likes it so for that case i made an option in frontend.

I can offer only two solutions:
1) leave the option as is and rename it to "SCALING" to make a bit less confusing.
2) remove the option but enable PS2 scaling by default(with proper radar/coronas).

erorcun:
what about 2nd option + combining PS2 scaling and PS2 hud coordinates into one define, and making proper radar/coronas into seperate define?

Fire-Head:
> what about 2nd option + combining PS2 scaling and PS2 hud coordinates into one define, and making proper radar/coronas into seperate define?

PC hud scaling in such case will remain broken and proper radar/coronas will be impossible without PS2_HUD.

> edit: I changed my mind, PS2 scaling define + PS2 hud define + seperate proper radar/coronas define all _without_ any frontend options, but PS2 scaling is enabled by default, is that ok?

If new defines are enabled by default(except PS2_HUD) and no one minds then ok.

ShFil119 (approved):
lgtm

PR #918, opened 2021-01-02
Add option to redirect output to file

ShFil119:
Can be useful for mobile ports.


Calinou:
PS: If this is not done already, remember to flush on every line printed (or on a regular time basis). Otherwise, logs won't be saved if the game crashes, making troubleshooting much harder.

madebr:
Diverting all logging calls to a function would be useful for ports.
This allows using the native android NDK logging facility: https://developer.android.com/ndk/reference/group/logging

src/skel/glfw/glfw.cpp
```diff
@@ -1488,6 +1488,10 @@ WinMain(HINSTANCE instance,
 int
 main(int argc, char *argv[])
 {
+#ifdef LOG_TO_FILE
+	freopen("re3_log.txt", "w", stdout);
+	freopen("re3_errors.txt", "w", stderr);
+#endif
```
erorcun, line 1494: Please move this one line below, so we can use it on WIN32.

PR #922, opened 2021-01-03
Update cmake script + add conan script + add gha CI

madebr:
Hello,

This pr:
- updates the cmake build script:
  * add option to make use of a vendored librw or librw binary package
  * fix `RE3_AUDIO` option
  * fix search of 3rd party libraries (mpg123, opusfile, Miles SDK)
  * create a binary package
  * add `.rc` file thus setting the executable icon
- bump the `vendor/librw` submodule
  * this is needed as it contains fixes added by https://github.com/aap/librw/pull/44
- add conanfile build script, this is used by a new Github Actions workflow to test on all relevant platforms

--- 
- gl3-sdl2 is built on github actions, but it currently does not build.
This error is known by the developers. So the error is currently ignored.
- ps2 is currently missing on the gha build matrix. You can @madebr me when this is wanted.


ShFil119:
Looks superb. I have two things in mind which could be changed.
- jobs for 32bit and 64bit, debug with release info build. Just glfw and openal. (it's hard to build SBCs)
- As we literally lack code for sdl2 in re3, can you comment them? After we gonna add at least something, we will turn it on. For now I think to be able easily check if build fails is more important.
btw I'm searching why windows-latest, d3d9, miles fails

madebr:
>     * jobs for 32bit and 64bit arm, debug with release info build. Just glfw and openal. (it's hard to build SBCs)

Ok, I'll look into this. The Julius repo you sent me yesterday gave me some ideas about caching.
Right now, I'm still looking into the Nintendo Switch as https://github.com/AGraber/re3-nx looks decent.

>     * As we literally lack code for sdl2 in re3, can you comment them? After we gonna add at least something, we will turn it on. For now I think to be able easily check if build fails is more important.

I commented out sdl2 and mss.

>       btw I'm searching why windows-latest, d3d9, miles fails

Here is a build log of a failing msvc+d3d9+mss: https://github.com/madebr/re3/runs/1641200129.
The errors are about `cAudioManager`.

ShFil119:
Can yo try on d3d9+mss build forcing 32bit mode? (as we only have this version of lib provided)

.github/workflows/build-cmake-conan.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: "Export Playstation 2 CMake toolchain conan recipe"
+        run: |
+          conan export vendor/librw/cmake/ps2toolchain ps2dev-cmaketoolchain/master@
+      - name: "Export librw conan recipe"
+        run: |
+          conan export vendor/librw librw/master@
+      - name: "Export Miles SDK conan recipe"
+        if: ${{ matrix.audio == 'miles' }}
+        run: |
+          conan export re3mss miles-sdk/master@
```
madebr, line 95: I have submitted a conan script to re3mss's repo to add a conan build script: see https://github.com/withmorten/re3mss/blob/master/conanfile.py
It is able to build a 64-bit mss import library.

As you can see from https://github.com/madebr/re3/runs/1641200129, the problem is about `cAudioManager`.
Maybe some define is missing for miles.

madebr: It's this `static_assert` that is firing:
https://github.com/GTAmodding/re3/blob/a6409fb44550870637ece43b0b5fe201ef5d7f1d/src/audio/AudioManager.h#L501-L503

madebr: 19220 = 4 * 4805
19220 = 8 * 2402.5

It's failing because the size of the `cAutioManager` is incorrect for 64-bit builds.

madebr: I'll look into 32-bit builds, but first I want to sort out caching because otherwise the 32-bit dependencies will have to be rebuilt each time.

ShFil119:
There's merge conflict. Maybe we should merge it and you would put extra stuff in next PR?

madebr:
I'll extract the merge conflict sensitive parts to another pr.
Then we can concentrate on the cmake/conan/ci things here.

ShFil119:
Do you think it's good example?
```
conan export vendor/librw librw/master@
mkdir build
cd build
conan install .. re3/master@ -if build -o re3:audio=openal -o librw:platform=gl3 -o librw:gl3_gfxlib=glfw --build missing -s re3:build_type=RelWithDebInfo -s librw:build_type=RelWithDebInfo
conan build .. -if build -bf build -pf package
```

madebr:
> Do you think it's good example?

The last step can be simplified a bit:
```
conan export vendor/librw librw/master@
mkdir build
cd build
conan install .. re3/master@ -if build -o re3:audio=openal -o librw:platform=gl3 -o librw:gl3_gfxlib=glfw --build missing -s re3:build_type=RelWithDebInfo -s librw:build_type=RelWithDebInfo
conan build ..
```

The `-s re3:build_type=RelWithDebInfo -s librw:build_type=RelWithDebInfo` is also not needed, but I added it to CI to get the pdb's.


Make sure that the user knows that he has to have re3 checked out, including submodules
```
git clone https://github.com/GTAmodding/re3.git
cd re3
git submodule init
git submodule update --recursive
```
Conan should also be installed: see https://conan.io/downloads.html
I use the one from pypi, but there exists a standalone exe for Windows.

madebr:
@ShFil119 
I also had to set `CONAN_SYSREQUIRES_MODE` to install the opengl system library:
https://github.com/GTAmodding/re3/blob/61036779140610c81688a636c3226c46571dbbef/.github/workflows/build-cmake-conan.yml#L100

PR #924, opened 2021-01-03
Entity/Physical proof-read 1

erorcun:
will be updated later

### DIFF
--- a/src/entities/Entity.cpp
+++ b/src/entities/Entity.cpp
@@ -126,6 +126,7 @@ CEntity::CreateRwObject(void)
 			m_matrix.AttachRW(RwFrameGetMatrix(RpAtomicGetFrame((RpAtomic*)m_rwObject)), false);
 		else if(RwObjectGetType(m_rwObject) == rpCLUMP)
 			m_matrix.AttachRW(RwFrameGetMatrix(RpClumpGetFrame((RpClump*)m_rwObject)), false);
+
 		mi->AddRef();
 	}
 }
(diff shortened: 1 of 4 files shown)


src/entities/Entity.cpp
```diff
@@ -122,10 +122,9 @@ CEntity::CreateRwObject(void)
 	if(m_rwObject){
 		if(IsBuilding())
 			gBuildings++;
-		if(RwObjectGetType(m_rwObject) == rpATOMIC)
-			m_matrix.AttachRW(RwFrameGetMatrix(RpAtomicGetFrame((RpAtomic*)m_rwObject)), false);
-		else if(RwObjectGetType(m_rwObject) == rpCLUMP)
-			m_matrix.AttachRW(RwFrameGetMatrix(RpClumpGetFrame((RpClump*)m_rwObject)), false);
+		if(RwObjectGetType(m_rwObject) == rpATOMIC || RwObjectGetType(m_rwObject) == rpCLUMP)
```
aap, line 125: i think this is a compiler optimization since both branches are the same. but since we're already checking object type i think we should use the typed functions.

src/entities/Entity.cpp
```diff
@@ -135,10 +134,9 @@ CEntity::AttachToRwObject(RwObject *obj)
 {
 	m_rwObject = obj;
 	if(m_rwObject){
-		if(RwObjectGetType(m_rwObject) == rpATOMIC)
-			m_matrix.Attach(RwFrameGetMatrix(RpAtomicGetFrame((RpAtomic*)m_rwObject)), false);
-		else if(RwObjectGetType(m_rwObject) == rpCLUMP)
-			m_matrix.Attach(RwFrameGetMatrix(RpClumpGetFrame((RpClump*)m_rwObject)), false);
+		if(RwObjectGetType(m_rwObject) == rpATOMIC || RwObjectGetType(m_rwObject) == rpCLUMP)
```
aap, line 137: same here.

aap (commented):
good stuff

PR #946, opened 2021-01-11
audio: only use #pragma comment(lib, xxx.lib) on MSVC

madebr:
`#pragma comment(lib "libmpg123-0.lib")` is only available on the MS C compiler.

This commit is extracted from #922 to avoid future merge conflicts.


madebr:
I just saw that `#pragma comment(lib, xxx.lib)` is used in multiple places.
But let's convert those other places when they become troublesome.

PR #951, opened 2021-01-12
Ci fixes

ShFil119:


madebr:
Was there a bug?

ShFil119:
This part, using internal vs not using at all.
https://github.com/GTAmodding/re3/commit/4fe64d1b0d96bf5fdf8e202932815cb5bb1d2678

PR #957, opened 2021-01-13
Fix build without FIX_BUGS, divide to 0 fixes

erorcun:


src/core/config.h
```diff
@@ -238,12 +238,16 @@ enum Config {
...
 #endif
 
-#define FIX_BUGS		// fixes bugs that we've came across during reversing, TODO: use this more
+#define FIX_BUGS		// fixes bugs that we've came across during reversing. You can undefine this only on release builds.
 //#define MORE_LANGUAGES		// Add more translations to the game
 #define COMPATIBLE_SAVES // this allows changing structs while keeping saves compatible
 #define LOAD_INI_SETTINGS // as the name suggests. fundamental for CUSTOM_FRONTEND_OPTIONS
 #define FIX_HIGH_FPS_BUGS_ON_FRONTEND
 
+#if defined(__LP64__) || defined(_WIN64)
```
ShFil119, line 247: Maybe we should add to premake/cmake new defines?

erorcun: right now this covers all our 64 bit configurations. if you know any platform that needs any other define than that, sure.

ShFil119 (approved):

PR #959, opened 2021-01-13
Lcs Particle

Fire-Head:


src/render/ParticleMgr.cpp
```diff
@@ -4,39 +4,63 @@
...
-	memset(this, 0, sizeof(*this));
+#ifdef FIX_BUGS
+	m_aParticles = nil;
+#endif
+}
+
+cParticleSystemMgr::~cParticleSystemMgr()
+{
+#ifdef FIX_BUGS
+	if ( m_aParticles )
```
Sergeanur, line 23: no if needed, second one too

PR #963, opened 2021-01-17
all LCS script commands

Nick007J:


src/core/config.h
```diff
@@ -211,7 +212,7 @@ enum Config {
 #		define PS2_MATFX
 #	endif
 #	define PC_PLAYER_CONTROLS	// mouse player/cam mode
-#	define GTA_REPLAY
+#	//define GTA_REPLAY
```
Sergeanur, line 215: Better to comment before the # maybe?

src/control/Script.h
```diff
@@ -549,7 +553,33 @@ class CRunningScript
...
+			return;
+		}
+		if (m_nAndOrState >= ANDS_1 && m_nAndOrState <= ANDS_8) {
+			m_bCondResult &= flag;
+			if (m_nAndOrState == ANDS_1) {
+				m_nAndOrState = ANDOR_NONE;
+				return;
+			}
+		}
+		else if (m_nAndOrState >= ORS_1 && m_nAndOrState <= ORS_8) {
```
x87, line 571: FYI: `ANDS_1` and `ORS_1` should be equal to 2 and 22 respectively
https://github.com/sannybuilder/dev/issues/56#issuecomment-678015657

src/control/ScriptCommands.h
```diff
@@ -1478,186 +1478,186 @@ enum {
...
+	COMMAND_IS_WORLD_STREAMING_COMPLETE,
+	COMMAND_SWAP_TO_STREAMED_SECTOR,
+	COMMAND_SET_CHAR_ATTACKS_PLAYER_WITH_COPS,
+	COMMAND_REGISTER_FACE_PLANT_DISTANCE,
+	COMMAND_REGISTER_MAX_SECONDS_ON_CARNAGE_LEFT,
+	COMMAND_REGISTER_MAX_KILLS_ON_RC_TRIAD,
+	COMMAND_REGISTER_HIGHEST_LEVEL_SLASH_TV,
+	COMMAND_ADD_MONEY_MADE_WITH_SLASH_TV,
+	COMMAND_ADD_TOTAL_KILLS_ON_SLASH_TV,
+	COMMAND_ADD_NOODLES_DELIEVERED,
```
x87, line 1604: typo

src/control/ScriptCommands.h
```diff
@@ -1478,186 +1478,186 @@ enum {
...
-	COMMAND_1557,
-	COMMAND_1558,
-	COMMAND_1559,
-	COMMAND_1560,
-	COMMAND_1561,
-	COMMAND_1562,
-	COMMAND_1563,
-	COMMAND_1564,
-	COMMAND_1565,
+	COMMAND_SET_DEBUG_MENU_ACTUVE,
```
x87, line 1496: typo

src/control/Script.h
Nick007J: This is a tricky one. I think that enum remained the same. But the ANDOR processor itself has changed, as it no longer adds 1 to the value it reads from SCM: https://github.com/GTAmodding/re3/pull/963/files#diff-335a0373684a54b6cba1386027737f4e12bca721b0b74865bca93dfbc90ba65cL4624 So what was internally "ANDOR 1" before resulted in m_nAndOrState equalling 2 (ANDS_2). Now same is achieved with "ANDOR 2".

PR #964, opened 2021-01-17
CModelInfo::GetModelInfoFromHashKey

majesticCoding:


aap:
Hm, you half implemented the moving of modelinfo into a non-static struct here. let's keep that out for the moment and do it later.

erorcun:
so, what is the status of this?

aap:
i'm closing it. will do it myself later. needs more changes than these here.

PR #982, opened 2021-01-23
Add always accurate PI

ShFil119:


withmorten:
could you also add this to the undefs in VANILLA_DEFINES?

src/core/config.h
```diff
@@ -488,4 +497,5 @@ enum Config {
 #undef FREE_CAM
 #undef RADIO_SCROLL_TO_PREV_STATION
 #undef BIG_IMG
+#undef ALWAYS_ACCURATE_PI
```
madebr, line 500: This `undef` won't do anything because the macros `PI_3_14` and `PI_3_1415` are already defined to `M_PI`.

withmorten: This way of doing vanilla defines is now "gone", so not a problem anymore!

PR #983, opened 2021-01-23
Fix plane / ccbb cheat flight pitch

ZLau92:
Un-invert the flight-stick for up-down steering for plane and flying cheat controls


erorcun:
is this original?

ZLau92:
this makes it behave like vanilla. I tried looking in ghidra to see if the negative is present, but I couldn't locate the FlyingControl function.

aap:
It *is* original...weird we never noticed this before...

Issue #837, opened 2020-11-25
GTA 3 1.1 Transparent textures are not visible from one side


Dima353:
Bellyup car problem, transparent body. 
At train stops there are transparent stairs and perryl, the roofs of stops.
game data files from Steam version.
![1](https://user-images.githubusercontent.com/75039669/100268715-73bf0400-2f66-11eb-9e82-8f50ec60cca0.PNG)
![2](https://user-images.githubusercontent.com/75039669/100268722-77528b00-2f66-11eb-97b6-d1e1cd27d744.PNG)
![3](https://user-images.githubusercontent.com/75039669/100268731-79b4e500-2f66-11eb-8bc6-ea95c4caece5.PNG)

aap:
That's a problem of the backface culling list. I stole it from android but it seems rather incomplete.

Dima353:
> Это проблема списка отбраковки обратной стороны. Я украл его с андроида, но он кажется неполным.

Is this problem solvable?

aap:
we'd have to find more bugged models and then the vita port would need a new release.

Dima353:
> we'd have to find more bugged models and then the vita port would need a new release.

Strange, but in version 1.0 for ps vita, there were no such problems

Dima353:
> we'd have to find more bugged models and then the vita port would need a new release.

Found more

![photo_2020-11-25_22-06-15](https://user-images.githubusercontent.com/75039669/100271405-7e7b9800-2f6a-11eb-996c-d417fb2eb2ac.jpg)

Dima353:
> we'd have to find more bugged models and then the vita port would need a new release.

Checked all cars defect only with Bellyup.

aap:
I have fixed the bug with the belly up and will push it soon. in the meantime feel free to find more buggy objects.

Dima353:
> I have fixed the bug with the belly up and will push it soon. in the meantime feel free to find more buggy objects.
 
Replaced the Bellyup texture with the XBOX version, the transparency is gone, the model is displayed well

Dima353:
> I have fixed the bug with the belly up and will push it soon. in the meantime feel free to find more buggy objects.

Found more
![photo_2020-11-27_00-21-15](https://user-images.githubusercontent.com/75039669/100392124-94f72180-3046-11eb-9856-b93988cc85da.jpg)
[...]

PR #986, opened 2021-01-24
Add toggle for (un)named semaphores and prevent linux header inclusion on non-linux paltforms

AGraber:
This adds a compile-time define that decides whether unnamed or named semaphores are used, since some platforms may not support one or the other.

Also prevents a Linux header from being included if we're not building for Linux.

This originally was a PR for the Nintendo Switch port but it's going to take a while, so let's just merge the 2 commits that are ready for now.

Original PR contents (for archival purposes):

----

This PR will bring Nintendo Switch fork loosely based on the old fork I did a couple months ago.

- [ ] CMake files to use devkitPro's Switch toolchain
- [x] Fix OpenAL variable clash when linking (#837)
- [x] Toggle (un)named semaphore usage on CdStreamPosix through ~~macro~~ function (named semaphore returns ENOSYS which means not implemented on the Switch)
- [ ] Check that the CdStreamPosix.cpp threads exit gracefully because you can't kill threads on the switch.
- [ ] Check wav support implementation (since it's using cmake, libsndfile should be easier to compile now and might be used instead of dr_wav on the old fork)
- [ ] Check audio extensions code (some EAX stuff triggers a crash on the openal switch port)
- [ ] Check `DIGITALRATE` macro for the Switch (on the old fork any value other than 48000 seems to crash)
- [ ] Check/swap X and circle/A and B controller mappings (since A on the switch is where Circle would be, but A is used for confirming rather than cancelling/going back)
- [ ] Check some custom config defines for the switch
- [ ] Check `realpath` which is apparently missing from the Switch's toolchain (?)
- [ ] Check casepath and path related functions to support switch mounted devices (paths that start with "[something]:/" like "sd:/switch/re3/models/GTA3.IMG")
- [ ] Ignore useless glfw input devices (touchscreen input implemented as "mouse" on glfw, some controller buttons getting passed as "keyboard" at the same time, etc.)
- [ ] Check screen resolution stuff (always has to be either 720p when docked, 1080p when docked)
- [x] ~~Switch stuff on main~~ glfw seems to handle this for us so no need to do that like on the old fork

Optional but nice-to-haves:
- [ ] Switch button prompts (nxbtns.txd on old fork's release)
- [ ] CPU Boost on loading screens (The switch has a feature which throttles the GPU but boosts the CPU clock up to 1,7GHz, meant for faster loading, actually saves around 3-4 seconds) examples: [1-1](https://github.com/AGraber/re3-nx/blob/e3a17f7fe35a6170eb561a58dc05631d27eb4af0/src/skel/glfw/glfw.cpp#L1556) [1-2](https://github.com/AGraber/re3-nx/blob/e3a17f7fe35a6170eb561a58dc05631d27eb4af0/src/skel/glfw/glfw.cpp#L1888) [2-1](https://github.com/AGraber/re3-nx/blob/e3a17f7fe35a6170eb561a58dc05631d27eb4af0/src/core/Game.cpp#L401) [2-2](https://github.com/AGraber/re3-nx/blob/e3a17f7fe35a6170eb561a58dc05631d27eb4af0/src/core/Game.cpp#L665)
- [ ] GPU throttle on pause menu (same as before but without the CPU boost, good for power saving)

### DIFF
--- a/src/core/CdStreamPosix.cpp
+++ b/src/core/CdStreamPosix.cpp
@@ -1,8 +1,8 @@
 #ifndef _WIN32
 #include "common.h"
 #include "crossplatform.h"
-#include <pthread.h>
 #include <signal.h>
+#include <pthread.h>
 #include <semaphore.h>
 #include <sys/types.h>
 #include <unistd.h>
@@ -13,7 +13,10 @@
 #include <fcntl.h>
 #include <sys/resource.h>
 #include <stdarg.h>
+
+#ifdef __linux__
 #include <sys/syscall.h>
+#endif
 
 #include "CdStream.h"
 #include "rwcore.h"
--- a/src/core/CdStreamPosix.cpp
+++ b/src/core/CdStreamPosix.cpp
@@ -12,6 +12,7 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <sys/resource.h>
+#include <stdarg.h>
 #include <sys/syscall.h>
 
 #include "CdStream.h"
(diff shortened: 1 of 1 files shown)


AGraber:
Regarding `pthread_kill` usage: I think we should be able to replace it with `pthread_cancel` without issues, that way we can also get rid of `#include <sys/syscall.h>`. Thoughts?

AGraber:
syscall.h header was actually used for some stuff under a `#ifdef __linux__`, so I'll do the same for the inclusion

AGraber:
Undid pthread_kill -> pthread_cancel replacement for now, after some investigation the Switch can't even kill threads, this will be checked later.

GaryOderNichts:
>  Check audio extensions code (some EAX stuff triggers a crash on the openal switch port)
I also had that issue on the Wii U port.
Loading the function pointers using `alGetProcAddress` instead of linking against them fixes it.
Since commit https://github.com/GTAmodding/re3/commit/5a1ea7bcf579b616e349ed651c67e0f988592cbd they also don't collide anymore.

PR #1003, opened 2021-01-27
Use original animation names from VCS

Sergeanur:


src/animation/AnimationId.h
```diff
@@ -2,264 +2,264 @@
...
-	ANIM_THROWABLE_THROW = ANIM_WEAPON_FIRE,
-	ANIM_THROWABLE_THROWU,
-	ANIM_THROWABLE_START_THROW,
-	ANIM_WEAPON_FIRE_2ND = ANIM_WEAPON_CROUCHFIRE,
-	ANIM_WEAPON_FIRE_3RD = ANIM_WEAPON_SPECIAL,
+	ANIM_ATTACK_1,
+	ANIM_ATTACK_2,
+	ANIM_ATTACK_EXTRA1,
+	ANIM_ATTACK_EXTRA2,
+	ANIM_ATTACK_3,
```
erorcun, line 242: seeing ANIM_ATTACK_EXTRA2 (which is idle anim in melee) in Walk/MoveAnim functions, ANIM_ATTACK_EXTRA1 as reload etc. is confusing imo. I'd like to use our synoyms for those, we can add some comment like "these are synonyms we added to clarify things" etc.

PR #1040, opened 2021-02-14
[Miami] Minor pass of the Spanish translation

IlDucci:
I wanted to do a full pass of the translation, but since I found out the issue with the opening exclamation mark in Pricedown and with the public announcement of re3/VC, I thought on importing some of the changes I'm planning to do right now.
 - Swapped all instances of an opening exclamation mark on strings that exclusively use the Pricedown font with a lowercase i to camouflage the fact.
 - Pasted a few strings from the current status of RE3's Spanish translation. These include mostly menus, but also include Donald Love's reference from GTA3 in Avery's Two Bit Hit mission (they were not in sync).
 - Added a few comments I used to know where I was. It's not all of them.
 - Other fixes and changes.
 - Lowercased some instances of non-RAE title uppercasing, like in the Colonel's name.
 - Who the heck thought "Sun Auto" was a "Solarium"?
 - Shortened "Invert First Person View" string, as it was touching the right-hand text.


Sergeanur:
If we're gonna replace reversed exclamation mark we should do it in the code, not in the text source.

IlDucci:
Code's not enough. Font1 lacks the opening exclamation mark for Pricedown, and we mentioned in the server that editing the font file was a no-go.

PR #1050, opened 2021-02-14
Fix seg fault when configuring controller

swills:
Since controllerAction can be -2 (line 2105)


swills:
Ok, did that. I'm confused how this could have been in the original code tho, this guarantees a crash when pulling up the controller config customization menu, but that didn't seem to happen. Maybe a compiler issue or something?

erorcun:
It doesn't guarantee a crash, it just guarantees on your configuration/system :) It's a typical array underflow. You can't believe how many overflow/underflows we fixed in the original code.

Thanks again, merging it.

PR #1052, opened 2021-02-14
Fix Lookright Driveby Anim (with passenger)

ZLau92:
Fixes driveby animation when looking right & passenger is in vehicle.


aap:
Hm, this is not original code and they quite deliberately didn't apply the anim if there's a passenger.

ZLau92:
let me double check this

PR #1070, opened 2021-02-16
Added INI option for max FPS.

sherief:


withmorten:
We're still debating on whether we want to add this - if we do it, we'll do it ourselves!

PR #1083, opened 2021-02-16
pc/ps2/custom limits

withmorten:
not done yet, waiting for input for what should be in limits_custom.h ...


thelink2012 (commented):
Not entirely sure on what should be increased too. @ThirteenAG should have more context due to P2DFX, but I think [III.VC.SA.LimitAdjuster.ini](https://github.com/ThirteenAG/III.VC.SA.LimitAdjuster/blob/master/doc/III.VC.SA.LimitAdjuster.ini) should have safe defaults (or [this one](https://github.com/ThirteenAG/III.VC.SA.LimitAdjuster/blob/v1.3/doc/limit_adjuster_gta3vcsa.ini) before unlimited was a thing). Fastman's ini should be helpful too.

ThirteenAG:
Well, I actually have no idea, p2dfx by itself doesn't require any limits to be changed, cuz corona rendering was replaced with buffered coronas by silent, and I set that limit internally only. Then, there's various draw distances that can be adjusted, I don't think p2dfx for re3 should have that feature, I think draw distances should be adjustable via menu or a separate mod.
Then, when you increase these draw distances, you probably need some map limits increased also, but since I don't know how the game works, I don't know which limits exactly. 
Talking strictly about coronas, for p2dfx, as far as I remember, gta sa has like 9k coronas with full map draw distance, III and VC probably have something like 2k and 4k (NUMCORONAS = 56 cries in the corner).

PR #1127, opened 2021-02-18
Update german.txt

RuesanG:
My english ist not 100% perfectly i give my best

### DIFF
--- a/utils/gxt/german.txt
+++ b/utils/gxt/german.txt
@@ -8031,28 +8031,28 @@ BILINEAR
 TRILINEAR
 
 [FED_WND]
-WINDOWED
+FENSTERMODUS
 
 [FED_FLS]
-FULLSCREEN
+VOLLBILD
 
 [FEM_CSB]
 CUTSCENE BORDERS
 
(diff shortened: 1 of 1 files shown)


utils/gxt/german.txt
```diff
@@ -8062,13 +8062,13 @@ FREE CAM
 
 { Linux joy detection }
 [FEC_JOD]
-DETECT JOYSTICK
+JOYSTICK ERKENNEN
 
 [FEC_JPR]
-Press any key on the joystick of your choice that you want to use on the game, and it will be selected.
+Drücken Sie eine beliebige Taste auf dem Joystick Ihrer Wahl, den Sie für das Spiel verwenden möchten, und es wird ausgewählt.
```
withmorten, line 8068: I'd say

`Drücken Sie eine beliebige Taste auf dem Joystick den Sie für das Spiel verwenden möchten, und es wird ausgewählt.`

Not sure if es or er ...

RuesanG: the way you write it, it fits me too, it sounds better than what I wrote

PR #1128, opened 2021-02-18
CMake: MSS / mpg123 compilation fixes

nick7:


ShFil119:
Good ones. Thanks. :)

PR #1134, opened 2021-02-19
German is 100% 

RuesanG:
German is 100% thanks to withmorten for your help  


utils/gxt/german.txt
```diff
@@ -8019,10 +8019,10 @@ GRAFIK-SETUP
 MIP MAPPING
 
 [FED_AAS]
-ANTI ALIASING
+KANTENGLÄTTUNG
```
withmorten, line 8022: I wouldn't translate this string, Anti Aliasing is actually also used in German this way.

RuesanG: since i have seen in many games that the anti-aliasing (Kantenglättung) was also there thought i would do it that way

withmorten: then it's fine.!

utils/gxt/german.txt
```diff
@@ -8031,44 +8031,44 @@ BILINEAR
...
+FREIE KAMERA
 
 { Linux joy detection }
 [FEC_JOD]
-DETECT JOYSTICK
+JOYSTICK ERKENNEN
 
 [FEC_JPR]
-Press any key on the joystick of your choice that you want to use on the game, and it will be selected.
+Drücken Sie eine beliebige Taste auf dem Joystick den Sie für das Spiel verwenden möchten, und er wird ausgewählt.
```
aeifr, line 8068: "```...und er wird ausgewählt....```"
"er"? Just a typo and you wanted to type "es" or "sie"(die Taste/the key(feminimum))?

withmorten: "Der" Joystick wird ausgewählt (bzw Controller), I think proofread this back then and thought it was right ...

aap: Doesn't game use "du" to address the user?

aeifr: @app 
Depends on the form you want to use. The respectful form would be "Sie", the simpler, more directive would be "du".
There should be more like a guideline or at least an advice for other languages which also offers two versions to address a single person. Just to be equally across all languages.

aeifr: @withmorten 
Joystick is fine, even for the german language. Iirc, even in the beginnng of the 21th century, joysticks where sold without any special fancy pancy translation. Just joystick. Codewise everything is hinting towards a joystick. Take a few examples of joystick related translation labels and look into the code. The context is directing to joysticks, not gamepads/controllers.

aap: @aeifr i just don't recall the game siezing me when i played it in german (it's been a while since i last did though)

aeifr: @withmorten 
Ouh, you are right. The context might be bound to the joystick, not the button. 
Sorry for the hassle. Took some time to realize. 

@aap 
Siezing. Needed some time to notice you just englishfied the german word "Sie". D'oh. 
Me too. So you're suggestion (implicitly) to use the simple version, just "du"zing (proper german term: duzen").

So the translation might look like:
```Drücke eine beliebige Taste auf dem Joystick den du für das Spiel verwenden möchtest, und er wird ausgewählt.```

RuesanG: Since I have translated to the best of my knowledge, since you are the developers, you can assess better because I prefer "Sie" than Du

withmorten: Fixed to "duzen"

PR #1135, opened 2021-02-19
Update reVC German.txt

RuesanG:
reVC German is 100% translated :)


utils/gxt/german.txt
```diff
@@ -14644,25 +14644,25 @@ MISSION WIEDERHOLEN
...
-VEHICLE PIPELINE
+FAHRZEUG-PIPELINE
 
 [FED_PRM]
-PED RIM LIGHT
+CHARAKTER KANTEN LICHT
 
 [FED_RGL]
-ROAD GLOSS
+GLÄNZENDE STRAßEN
```
aeifr, line 14653: There is an uppercase of ß - ẞ in the german language (minor visual differences)
GLÄNZENDE STRAẞEN

Alternativ translation suggestion:
Straßenglanz

withmorten: That doesn't work in GTA (just tried and it just freezes).

PR #1136, opened 2021-02-19
fix premake5.lua for macOS build

peternguyen93:
- fix homebrew path in macOS amd64 and arm64 to /opt/homebrew
- seperate CFLAGS and CXXFLAGS while generating gmake2 with premake5

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -137,12 +137,18 @@ workspace "reVC"
 	filter { "platforms:*arm*" }
 		architecture "ARM"
 		
-	filter { "platforms:macosx-arm64-*" }
-		buildoptions { "-target", "arm64-apple-macos11", "-std=gnu++14" }
-
-	filter { "platforms:macosx-amd64-*" }
-		buildoptions { "-target", "x86_64-apple-macos10.12", "-std=gnu++14" }
-
+	filter { "platforms:macosx-arm64-*", "files:**.cpp"}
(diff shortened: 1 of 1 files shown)


erorcun:
you don't like X11 i see

premake5.lua
```diff
@@ -342,15 +348,15 @@ project "reVC"
 		libdirs { "vendor/openal-soft/libs/Win64" }
 
 	filter "platforms:linux*oal"
-		links { "openal", "mpg123", "sndfile", "pthread", "X11" }
```
erorcun, line 345: i assume this and below one were unintended

premake5.lua
```diff
@@ -204,9 +210,9 @@ project "librw"
 
 	filter "platforms:macosx*"
 		-- Support MacPorts and Homebrew
-		includedirs { "/opt/local/include" }
+		includedirs { "/opt/homebrew/include" }
 		includedirs {"/usr/local/include" }
-		libdirs { "/opt/local/lib" }
+		libdirs { "/opt/homebrew/lib" }
```
erorcun, line 215: i think it will be better to keep old paths, and add `/opt/homebrew` ones as extra

erorcun:
/usr/local should stay for AMD64 Macs, and /usr/homebrew should be enabled only for ARM64 Macs. I'll do it myself, but thanks for the PR.

Closing.

PR #1160, opened 2021-06-27
Fix move direction while shooting and moving

yasinkaraaslan:
Move direction was broken when shooting and moving with Colt45 and UZI (only in Free Camera mod), fixed it. The code is taken from VC.


yasinkaraaslan:
I accidentally deleted some stuff. Sorry

PR #1161, opened 2021-06-27
Fix move direction while shooting and moving

yasinkaraaslan:
Move direction was broken when shooting and moving with Colt45 and UZI (only in Free Camera mod), fixed it. The code is taken from VC.

### DIFF
--- a/src/peds/PlayerPed.cpp
+++ b/src/peds/PlayerPed.cpp
@@ -510,6 +510,25 @@ CPlayerPed::DoWeaponSmoothSpray(void)
 	return false;
 }
 
+float
+CPlayerPed::GetWeaponSmoothSprayRate(void)
+{
+	if(m_nPedState == PED_ATTACK && !m_pPointGunAt) {
+		CWeaponInfo *weaponInfo = CWeaponInfo::GetWeaponInfo(GetWeapon()->m_eWeaponType);
+		switch(GetWeapon()->m_eWeaponType) {
+		case WEAPONTYPE_SHOTGUN:
+		case WEAPONTYPE_AK47:
+		case WEAPONTYPE_M16: return PI / 180.f;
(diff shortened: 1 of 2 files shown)


erorcun:
Please read the PR template/instructions.

yasinkaraaslan:
It's a game bug fix, what's wrong with it? It's not a fix to R* code because free camera isn't available in vanilla GTA 3. I was just testing the game and noticed this bug, tried to fix it and thought it would be nice to share.
If I done something wrong I am sorry.

erorcun:
You didn't read it completely.

> We don't accept custom codes, **as long as it's not wrapped via preprocessor conditions**, or it's linux/cross-platform skeleton/compatibility layer.

You should use FREE_CAM in here + make sure free cam is activated for what you want to achieve.

yasinkaraaslan:
This function is not exclusive to the free camera. There is no change in the camera at all, just a minor fix for the gameplay. The code was working great with the old camera but it had problems with the new one, that's why I mentioned it. **m_fRotationDest** variable is more accurate now, that's all, which has nothing to do with the camera at all. Just movement direction. It's not a big deal. 

It was a headache to find the code responsible for this though...

erorcun:
>... all of the code on the repo that's not behind a preprocessor condition(like FIX_BUGS) are **completely reversed code** from original binaries.

Is it too hard to understand this?

yasinkaraaslan:
Whatever... I was just trying to help. Good luck, great project

yasinkaraaslan:
Isn't this the same thing? You did the same thing for reVC and that's where I took the code from. You didn't wrap it or anything like that it seems.

https://github.com/GTAmodding/re3/commit/bd8b907d131ac9e5b471a0a31928849b95f5d1f0

PR #1188, opened 2021-07-01
fix: printHash.sh; replace echo -n, check if $1 is set.

mssx86:
this pull request rewrites `printHash.sh` to get rid of the dependency on bash and bashisms (the use of `echo -n`) for hosts without bash. it also does a simple check to see if `$1` is set and removes the unnecessary `>$1` as the first invocation of `printf` already does that.

checked with `bash 5.1.8`, `yash 2.51` and `busybox ash 1.33.1`.

### DIFF
--- a/printHash.sh
+++ b/printHash.sh
@@ -1,12 +1,14 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
+if [ -z "${1}" ]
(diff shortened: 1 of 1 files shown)


withmorten:
cherry-picked them both over to miami!

PR #1189, opened 2021-07-01
Add Max FPS option in .ini file

leanndroguedes:


withmorten:
We're still debating on whether we want to add this - if we do it, we'll do it ourselves!

PR #1195, opened 2021-07-02
Fix move direction with certain weapons

yasinkaraaslan:
When moving and shooting with UZI and COLT45 (Only in Free Camera), the player's move direction was broken, fixed it. I took the code from VC and made some changes myself.

### DIFF
--- a/src/peds/Ped.cpp
+++ b/src/peds/Ped.cpp
@@ -1321,9 +1321,11 @@ CPed::CalculateNewVelocity(void)
 			limitedRotDest -= 2 * PI;
 		}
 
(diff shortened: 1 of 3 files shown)


erorcun:
If it's a "bug" created by Free Cam, you have to wrap it with FREE_CAM. Why you insist on not accepting/understanding this?

yasinkaraaslan:
This also fixes some minor problems with general gameplay along with free camera fixes (Like having both flamethrower and colt45 same rotation change when moving).  Also this isn't wrapped with FREE_CAM in miami branch, it has FIX_BUGS already.

PR #1204, opened 2021-07-04
Add Nintendo Switch support

AGraber:
This PR will add Nintendo Switch support to re3:

- Uses GTA_SWITCH macro as recommended
- Add GitHub Actions workflow for the Switch
- Add support for `sd:/` and similar path schemes
- Add macro to ignore mouse/keyboard input
- Had to enable extensions on Switch - more on that later

Some stuff that needs to be fixed:
- [x] Fix OpenAL linking - for some reason, `find_package(OpenAL REQUIRED)` succeeds (or at least it doesn't error) but the cmake variables related to OpenAL appear to be empty and nothing gets linked.

Some stuff that I want to get finished before merging:
- [x] Assets for gamepad menu for Switch - this is already finished on old fork, I need to port it over to this PR
- [x] Rumble support through libnx - finished on old fork
- [x] Port over `librw_platform_target` equivalent
- ~~CPU boost for loading screens~~

Regarding extensions, we spent a lot of time on Discord trying to figure out what was the deal with extensions but ultimately there seems to be a lot of inconsistency regarding which extensions are enabled and not, to the point we don't know what is considered an extension or not. Some extensions on Linux are enabled even with extensions turned off on CMake, while these are disabled on the Switch toolchain (eg. strncasecmp, alloca, etc.), along a bunch of other functions. Ultimately, I think it's best to just enable them on Switch, and just use what works on Linux with those disabled with this configuration.


src/core/config.h
```diff
@@ -322,7 +322,7 @@ enum Config {
 //#	define PS2_MENU_USEALLPAGEICONS
 #else
 
-#	ifdef XINPUT
+#	if defined(XINPUT) || defined(GTA_SWITCH)
 #		define GAMEPAD_MENU		// Add gamepad menu
```
erorcun, line 326: controller presets and vibration are unavailable on GLFW input, so enabling this page will confuse users a lot

presets may work at the start, but try changing bindings

AGraber: presets seem to work, and I just added platform specific functions in the next commit.
what do you mean with changing bindings from keyboard setup? that works too but I'm not quite sure how they work along presets

AGraber:
@madebr tagging in case you want to help out clean up cmake stuff and maybe help looking into that openal bug

src/CMakeLists.txt
```diff
@@ -120,13 +120,19 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     )
 endif()
 
+if(NINTENDO_SWITCH)
+    set(${PROJECT}_C_CXX_EXTENSIONS ON)
```
madebr, line 124: What gnu extension do you need?

src/CMakeLists.txt
```diff
@@ -140,3 +146,33 @@ if(${PROJECT}_INSTALL)
         install(FILES $<TARGET_PDB_FILE:${EXECUTABLE}> DESTINATION "." OPTIONAL)
     endif()
 endif()
+
+# Build Nintendo Switch binaries
+if(NINTENDO_SWITCH)
```
madebr, line 151: Can you have a look at https://github.com/aap/librw/pull/75 and see how I abstracted away the special handling in `librw_platform_target`?
I think it can also be used here.

src/core/common.h
```diff
@@ -11,6 +11,10 @@
 #define __STDC_LIMIT_MACROS // so we get UINT32_MAX etc
 #endif
 
+#ifdef GTA_SWITCH
+#include <switch.h>
```
madebr, line 15: Advantages of this:
- `__SWITCH__` is defined by the switch toolchain (see https://github.com/devkitPro/pacman-packages/blob/c4c328a2ce70a8fbeb25a3833b3a0c285a5023ef/cmake/switch/Switch.cmake#L8)
- No need to add `GTA_SWITCH` in the cmake script.
Disadvantage:
- Always need to have `common.h` included when needing to check existence of `GTA_SWITCH`.
But I think this is acceptable. Or other uses should check for `__SWITCH__` instead (as they do for `_WIN32` and `__APPLE__`)

```suggestion
#ifdef __SWITCH__
#define GTA_SWITCH
#include <switch.h>
```

src/skel/glfw/glfw.cpp
```diff
@@ -324,6 +326,79 @@ psNativeTextureSupport(void)
...
+	HidInitializationGCResult = hidInitializeVibrationDevices(&SwitchVibrationDeviceGC, 1, HidNpadIdType_No1, HidNpadStyleTag_NpadGc);
+	if(R_FAILED(HidInitializationResult[1])) {
+		printf("Failed to initialize VibrationDevice for GC Mode\n");
+	}
+
+	SwitchVibrationValues[0].freq_low  = 160.0f;
+	SwitchVibrationValues[0].freq_high = 320.0f;
+
+	padConfigureInput(1, HidNpadStyleSet_NpadFullCtrl);
+    padInitializeDefault(&SwitchPad);
```
madebr, line 363:
```suggestion
	padInitializeDefault(&SwitchPad);
```

src/skel/glfw/glfw.cpp
```diff
@@ -324,6 +326,79 @@ psNativeTextureSupport(void)
...
+static HidVibrationValue SwitchVibrationValues[2];
+static HidVibrationDeviceHandle SwitchVibrationDeviceHandles[2][2];
+static HidVibrationDeviceHandle SwitchVibrationDeviceGC;
+
+static PadState SwitchPad;
+
+Result HidInitializationResult[2];
+Result HidInitializationGCResult;
+
+void _psInitializeVibration()
```
madebr, line 344:
```suggestion
static void _psInitializeVibration()
```

src/skel/glfw/glfw.cpp
```diff
@@ -324,6 +326,79 @@ psNativeTextureSupport(void)
...
+	}
+
+	SwitchVibrationValues[0].freq_low  = 160.0f;
+	SwitchVibrationValues[0].freq_high = 320.0f;
+
+	padConfigureInput(1, HidNpadStyleSet_NpadFullCtrl);
+    padInitializeDefault(&SwitchPad);
+}
+
+void _psHandleVibration()
```
madebr, line 366:
```suggestion
static void _psHandleVibration()
```

src/CMakeLists.txt
madebr: Nevermind, it's in your initial post. Sorry.

madebr:
> @madebr tagging in case you want to help out clean up cmake stuff and maybe help looking into that openal bug

The openal package of devkitpro contains a cmake configuration script.
After configuration, the cmake cache contains `OpenAL_DIR:PATH=/opt/devkitpro/portlibs/switch/lib/cmake/OpenAL`.
In that directory, the file `OpenALConfig.cmake` defines the `OpenAL::OpenAL` target directly.
So no `find_file`/`file_library` is performed in `find_package`, which causes no openal related cached variabled to appear in the cmake cache.

AGraber:
> So there's no openal bug. The only problem might be that it violated the api at [`FindOpenAL.cmake`](https://cmake.org/cmake/help/latest/module/FindOpenAL.html):
> `OPENAL_INCLUDE_DIR`, `OPENAL_LIBRARY` and `OPENAL_VERSION_STRING` are not defined.
> But we don't use those variables.

They're actually used:
https://github.com/GTAmodding/re3/blob/ab7513cc85e980a796a80950af1d06929693adc1/src/CMakeLists.txt#L55-L61

With your explanation it makes sense that the variables are then empty and nothing gets linked. We'll have to wait until it's fixed on their side before removing the small `target_link_libraries` hack then.

As for the `GTA_SWITCH` macro, I think it's better to just use the `__SWITCH__` macro instead since other platforms already do it like that as you mentioned (`__APPLE__`, etc.)

madebr:
> They're actually used:

Yes, I spoke too soon.

I fixed the problem on https://github.com/madebr/re3/tree/pr1204 in a *clean* way.
I believe the problem was introduced by https://github.com/devkitPro/pacman-packages/pull/241#issuecomment-869077864.

Before, the `FindOpenAL.cmake` cmake module of cmake would end up being used.
After https://github.com/devkitPro/pacman-packages/commit/d9f8148c77c09023e1bffb333ef125c40307e8c8, cmake will prefer the config script to the find script.

It can also be fixed by using `find_package(OpenAL REQUIRED MODULE)`.
Perhaps that's the nicest way?

AGraber:
The 2nd is probably the nicest since `OpenAL::OpenAL` isn't present in any official cmake documentation/implementation. That said, both sound like good options but I think it's a good idea to try to get this fixed on devkitPro's side,

AGraber:
I've taken a closer look at how the OpenAL config is generated. The config file with the `OpenAL::OpenAL` variables is something provided by the OpenAL package itself (also in other systems). So, it can be said that `OpenAL::OpenAL` would be way OpenAL intends you to use it, while the `OPENAL_XXX` variables are something provided by CMake. From some googling around, some repositories that provide OpenAL also provide a config similar to the one of devkitPro. With that said I think the 1st solution you already wrote is better, since it would favor the configuration that the library itself provides in some systems.

madebr:
I've created a pr on openal-soft's repo at https://github.com/kcat/openal-soft/pull/581 to make their generated `OpenALConfig.cmake` compatible with `FindOpenAL.cmake` from cmake.
Let's see how it is received.

AGraber:
Force pushed because the build system had some url to be updated (rip bintray)

src/core/config.h
erorcun: i think you will disable keyboard bindings menu?

AGraber: Yes, now the options menu goes directly to gamepad menu (through `GTA_HANDHELD` macro) cause the rest of the menu ends up being gamepad settings, restore settings and go back, which is kinda useless.

src/CMakeLists.txt
```diff
@@ -140,3 +155,28 @@ if(${PROJECT}_INSTALL)
...
+
+    nx_generate_nacp (${EXECUTABLE}.nacp
+        NAME    "${EXECUTABLE}"
+        AUTHOR  "${EXECUTABLE} Team"
+        VERSION "1.0.0-${GIT_SHA1}"
+    )
+
+    nx_create_nro(${EXECUTABLE}
+        NACP ${EXECUTABLE}.nacp
+        ICON "${PROJECT_SOURCE_DIR}/logo_switch.jpg"
```
madebr, line 174: I think we should group all resources in e.g. a `res/images` folder.
Also, instead of `logo_switch.jpeg` I would make the suffix the resolution.
So the tree structure would become:
```
res/images/
- logo.svg
- logo_1024.png
- logo_256.jpg
```
Does Nintendo switch only support the jpeg file type?

AGraber: Yeah, I was thinking the same as eventually the root directory will end up cluttered.

nx-hbmenu (the main way to load the game for now) only supports jpeg for the NRO icon, so I have to go with that.

adamboutcher:
Do we have an ETA for this merge to go mainline and a release compiled?

src/CMakeLists.txt
AGraber: Ported over as `re3_platform_target`. I would've used `${PROJECT}_platform_target` but cmake doesn't let you use variables to call commands

src/core/common.h
AGraber: Resolved by introducing new `GTA_HANDHELD` define and using `__SWITCH__` on platform specific things only.

AGraber:
Rebased with master, had to force-push.

> 
> 
> Do we have an ETA for this merge to go mainline and a release compiled?

Please be patient. I'm working this aside with other personal projects on my free time which fluctuates a lot because of studies, so I can't give any estimates. That said, it shouldn't be long since the main stuff is done.

AGraber:
Pretty much ready for review.

I'll explain some stuff that might not be clear by just seeing the diff:
## Switch controller
The Nintendo Switch Pro controller was added alongside the PS and XBOX ones. Originally intended for the Switch only, the Switch
Pro controller being a current-gen controller that is widely supported and used, I think this addition should be welcome into
re3 as a whole.

## `IGNORE_MOUSE_KEYBOARD` define
Some glfw ports emulate mouse/keyboard with the gamepad, which is undesireable in some build configurations as re3 already
supports gamepad properly. Only `GTA_HANDHELD` defines this for now.

## `GTA_HANDHELD` define
This define will set some things for handheld platforms, which includes the Nintendo Switch for now but should be used to
support other platforms in the future. This define enables the gamepad menu, defines `IGNORE_MOUSE_KEYBOARD`, disables PC
controls, the resolution selector, "video mode" (window or fullscreen) selector, and goes straight into the gamepad menu
in the options menu rather than going into the "controller menu" which would be relatively empty (since there's only stuff
like rebinds for mouse/keyboard and other stuff relevant to PC, which wouldn't work serve a purpose in `GTA_HANDHELD` builds).

## Github Workflow changes
New GitHub workflow that uses devkitPro's cmake to build the game for the Switch

## CMake changes
Some changes were made to accomodate for the Switch port, and potential future ports that might need a "platform target" build.
The build now also accounts for both CMake's and OpenAL's variables/aliases.

## `.vscode` config changes
Just fixes some previously incorrect configuration and adds intellisense configuration for building the Switch.

src/CMakeLists.txt
madebr: Imho, this `${PROJECT}` prefix is making the cmake script unreadable.
I would just keep the prefix `RE3` for all.

madebr: In a future refactoring, perhaps we can share the platform code between librw and re3.
It would make maintaining much easier.

src/skel/glfw/glfw.cpp
```diff
@@ -485,11 +562,13 @@ psInitialize(void)
 	debug("Physical memory size %llu\n", _dwMemAvailPhys);
 	debug("Available physical memory %llu\n", size);
 #else
+#ifndef __SWITCH__
  	struct sysinfo systemInfo;
 	sysinfo(&systemInfo);
 	_dwMemAvailPhys = systemInfo.freeram;
```
madebr, line 568: The physical memory size can be calculated using `HW_MEMSIZE` as `sys/sysctl.h` is available for switch.
`sysctl((int[]){CTL_HW, HW_PAGESIZE}, 2, &page_size, &ull_len, NULL, 0);` (and `HW_MEMSIZE`), that is already done for apple, is possible for switch.

The only thing that I'm unaware of, is a way to determine the available/free memory.

AGraber: `HW_MEMSIZE` is missing from the Switch's `sysctl.h` header.

I think this can be reimplemented with libnx functions instead.

madebr: devkitpro64 uses newlib, which has `mallinfo`.
Does that give you more information?
https://github.com/eblot/newlib/blob/2a63fa0fd26ffb6603f69d9e369e944fe449c246/newlib/libc/include/malloc.h#L77-L83
and
https://github.com/eblot/newlib/blob/2a63fa0fd26ffb6603f69d9e369e944fe449c246/newlib/libc/include/malloc.h#L22-L33

AGraber: `mallinfo` returns values that appear to be garbage or at least not useful for what's wanted here, claiming there's dozens of unmapped gigabytes of memory and etc.

On the other hand, libnx can only return the amount of memory the current application can use, but doesn't give any means to know how much it's actually using or how much of it it's free (it will report that there's a few mb free which is reserved for system stuff).

The best fit I can think of for now to ensure correct behavior in the future (since that `_dwMemAvailPhys` variable seems to be used for some stuff, albeit rarely) is saving how much memory the application can use into `_dwMemAvailPhys`.

src/skel/glfw/glfw.cpp
```diff
@@ -484,6 +561,9 @@ psInitialize(void)
 	_dwMemAvailPhys = (uint64_t)(vm_stat.free_count * page_size);
 	debug("Physical memory size %llu\n", _dwMemAvailPhys);
 	debug("Available physical memory %llu\n", size);
+#elif defined (__SWITCH__)
+	svcGetInfo(&_dwMemAvailPhys,     7, 0xffff8001, 0);
```
madebr, line 565: https://switchbrew.github.io/libnx/svc_8h.html
```suggestion
	svcGetInfo(&_dwMemAvailPhys,     7, CUR_PROCESS_HANDLE, 0);
```

src/skel/glfw/glfw.cpp
madebr: `malloc` is able to overcommit, so that is probably what `mallinfo` is reporting to you.
It's useful information to know how much you allocated, but not useful here I think.

src/skel/glfw/glfw.cpp
AGraber: Yeah, these are the intended defines, thanks. I actually grabbed the values from the switchbrew wiki but didn't knew they were available as defines already.

adamboutcher:
Thanks @AGraber your work is really appreciated! Also thanks for the update.

src/CMakeLists.txt
```diff
@@ -120,13 +124,24 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
...
+if(NINTENDO_SWITCH)
+    set(${PROJECT}_C_CXX_EXTENSIONS ON)
+else()
+    set(${PROJECT}_C_CXX_EXTENSIONS OFF)
+endif()
+
+if(NOT (WIN32 OR APPLE OR NINTENDO_SWITCH))
+    find_package(X11 REQUIRED)
+    target_link_libraries(${EXECUTABLE} PRIVATE X11::X11)
+endif()
```
erorcun, line 136: is this needed? because GTA_HANDHELD doesn't use GET_KEYBOARD_INPUT_FROM_X11 anyway, and I plan to fix this in a seperate commit.

madebr: I added this while helping @AGraber.
It's a fix for building re3 on Linux.
Not linking to X11 causes the following linking error:
```
[100%] Linking CXX executable re3
/usr/bin/ld: CMakeFiles/re3.dir/skel/glfw/glfw.cpp.o: undefined reference to symbol 'XGetKeyboardMapping'
/usr/bin/ld: /usr/lib64/libX11.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

erorcun: ah i see, but i was planning to do this in a seperate commit, by dynamically detecting if X11 exists(like compiling a dummy code that includes X11) and including it if so. So I suggest Graber to remove this part, and i will try doing it seperately, and i will ask for your review in that PR too.

src/CMakeLists.txt
```diff
@@ -120,13 +124,24 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
...
+
 set_target_properties(${EXECUTABLE}
```
madebr, line 138: @AGraber

Don't link to X11. This will get fixed soon by @erorcun.
```suggestion
endif()

set_target_properties(${EXECUTABLE}
```

PR #1205, opened 2021-07-05
Miami highfps: fix many visual issues and a few small physics issues at high framerates

Veyrdite:
Req merge in Veyrdite/re3:miami_highfps -> GTAmodding/re3:miami.

All changes wrapped in FIX_BUGS, using a variety of CTimer:: values where most appropriate.  

Bugs fixed:
 - ocean appearance at high FPS
 - particle appearance+physics at high FPS (including fountains & small black things that whirl around treetops)
 - boat throttle & steering ramp rate at high FPS
 - lots of cases where particles are created once per frame (spam at high FPS)
 - explain some boat physics code CBoat::ApplyWaterResistance() & FIX_BUGs a minor issue in it

Bugs NOT fixed (todo):
 - slow boats at high FPS (current project, "because surely the code is simpler than the cars", hah!)
 - bad car handling at high FPS
 - engine rev SFX ramp rates at high FPS (all boats & the top/overdrive gear on cars) 
 - many other cases where particles are created once per frame (spam at high FPS).  Rain, wave sparkles, ...

Features added:
 - Pressing F4 at any time during gameplay toggles the framelimiter (added to debug menu code).  Ridiculously useful for both fixing FPS related bugs and even to play the game (esp water missions, often the cutscenes softlock due to bad boat performance).

The framelimiter commit is a little bit messy and does not have any text message pop up on hud (dang uint16 strings!) or play a sound.  It's solely in debug code, however, and I think it's utility is very high.  Suggestions welcome for how I can implement it in a tidier manner.

### DIFF
--- a/src/extras/debugmenu.cpp
+++ b/src/extras/debugmenu.cpp
@@ -6,6 +6,7 @@
 #include "Timer.h"
 #include "rtcharse.h"
 #include "re3_inttypes.h"
+#include "Frontend.h"
 #include "debugmenu.h"
 #include <new>
 
@@ -1017,6 +1018,8 @@ DebugMenuProcess(void)
 	CPad *pad = CPad::GetPad(0);
 	if(CTRLJUSTDOWN('M'))
 		menuOn = !menuOn;
+	if (KEYJUSTDOWN(rsF4))
+		FrontEndMenuManager.m_PrefsFrameLimiter = !FrontEndMenuManager.m_PrefsFrameLimiter;
 	if(!menuOn)
 		return;
 
--- a/src/objects/ParticleObject.cpp
+++ b/src/objects/ParticleObject.cpp
@@ -423,6 +423,12 @@ CParticleObject::RemoveObject(void)
 void
 CParticleObject::UpdateAll(void)
 {
+#ifdef FIX_BUGS
+	// Fix particle generation spam at high FPS
+	if (CTimer::GetLogicalFramesPassed() == 0)
+		return;
+#endif
+
 	{
 		CParticleObject *pobj = pCloseListHead;
 		CParticleObject *nextpobj;
--- a/src/render/Particle.cpp
+++ b/src/render/Particle.cpp
@@ -1259,8 +1259,11 @@ void CParticle::Update()
 			}
 			
 			vecPos += vecMoveStep;
-			
+#ifdef FIX_BUGS
+			if ( psystem->m_Type == PARTICLE_FIREBALL && CTimer::GetLogicalFramesPassed()) // Fix particle spam at high FPS
+#else
 			if ( psystem->m_Type == PARTICLE_FIREBALL )
+#endif
 			{
 				  AddParticle(PARTICLE_HEATHAZE, particle->m_vecPosition, CVector(0.0f, 0.0f, 0.0f),
 					nil, particle->m_fSize * 5.0f);
@@ -1326,11 +1329,18 @@ void CParticle::Update()
 					
 					fDistToCam = (TheCamera.GetPosition() - vecPos).Magnitude();
 				}
-
+#ifdef FIX_BUGS
+				if ( numWaterDropOnScreen < nMaxDrops && numWaterDropOnScreen < 63
+					&& fDistToCam < 10.0f
+					&& clearWaterDrop == false
+					&& !CGame::IsInInterior() 
+					&& CTimer::GetLogicalFramesPassed()) // Fix waterdrop spam at high FPS
+#else
 				if ( numWaterDropOnScreen < nMaxDrops && numWaterDropOnScreen < 63
 					&& fDistToCam < 10.0f
 					&& clearWaterDrop == false
 					&& !CGame::IsInInterior() )
+#endif
 				{
 					CVector vecWaterdropTarget
 					(
@@ -1411,7 +1421,11 @@ void CParticle::Update()
 				}
 			}
 			
+#ifdef FIX_BUGS
+			if ( !(psystem->Flags & SCREEN_TRAIL) && CTimer::GetLogicalFramesPassed()) // Fix particle over-expansion at high FPS
+#else
 			if ( !(psystem->Flags & SCREEN_TRAIL) )
+#endif
 			{
 				float size;
 
@@ -1701,6 +1715,12 @@ void CParticle::Update()
 				}
 			}
 
+#ifdef FIX_BUGS
+			// Keep particles animating, rotating, fading etc at the right speed
+			// at high or low FPS.
+			for (uint32 i=0; i<CTimer::GetLogicalFramesPassed(); i++)
+#endif
+			{ // -- start FPS fix ---
 			if ( particle->m_nFadeToBlackTimer != 0 )
 			{
 				particle->m_nColorIntensity = Clamp(particle->m_nColorIntensity - particle->m_nFadeToBlackTimer,
@@ -1764,6 +1784,7 @@ void CParticle::Update()
 #else
 				particle->m_nRotation += particle->m_nRotationStep;
 #endif
+			} // -- end FPS fix --
 			
 			if ( particle->m_fCurrentZRadius != 0.0f )
 			{
@@ -1773,6 +1794,16 @@ void CParticle::Update()
 				
 				float fY = (Sin(nSinCosIndex) + Cos(nSinCosIndex)) * particle->m_fCurrentZRadius;
 
+#ifdef FIX_BUGS
+				// Prevent super-fast movement at high FPS
+				// We can use GetTimeSetpFix() here instead of GetLogicalFramesPassed()
+				// so the visual result looks "smooth" on player's screens.  We can't
+				// do this with most of the above Timers because the effects are stored
+				// as uint16's instead of floats, which means tiny changes each frame
+				// are often rounded down to zero change each frame.
+				fX *= CTimer::GetTimeStepFix();
+				fY *= CTimer::GetTimeStepFix();
+#endif
 				vecPos -= particle->m_vecParticleMovementOffset;
 
 				vecPos += CVector(fX, fY, 0.0f);
--- a/src/render/ParticleType.h
+++ b/src/render/ParticleType.h
@@ -81,7 +81,7 @@ enum tParticleType
 	PARTICLE_TEST,
 	PARTICLE_BIRD_FRONT,
 	PARTICLE_SHIP_SIDE,
-	PARTICLE_BEASTIE,
+	PARTICLE_BEASTIE,  // "Beasties" look like tree leaves circling up at canopy height, but are also used in other places (eg docks)
 	PARTICLE_RAINDROP_2D,
 	PARTICLE_HEATHAZE,
 	PARTICLE_HEATHAZE_IN_DIST,
--- a/src/render/WaterLevel.cpp
+++ b/src/render/WaterLevel.cpp
@@ -1968,7 +1968,7 @@ CWaterLevel::RenderWavyMask(float fX, float fY, float fZ,
 #ifndef PC_WATER
 			if (maskMorphVerts[base].z >= fMinSparkZ)
 #else
-			if ( maskMorphVerts[base].z > fMinSparkZ )
+			if (maskMorphVerts[base].z >  fMinSparkZ)
 #endif
 			{
 				switch ( (i + j + randval) & 3 )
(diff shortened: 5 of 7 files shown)


majesticCoding (approved):

Veyrdite:
Addendum: now slowly fixing vehicle physics.  I thought boats would be easier, turns out I accidentally found one of the causes of bad vehicle turning performance at high FPS instead.  Woops.

179450c is the most dramatic change that will probably have effects on other parts of the game too (helis?), too much rotational friction was being applied to all physical entities at high framerates.

Veyrdite:
I believe I've fixed all car handling related problems now.  I'm open if anyone thinks they have found an inaccuracy (ie cars & bikes not "feeling" right to play in some way).

ZLau92:
Do the car physics related changes fix the issue where you can push a car very easily on foot at high fps? Is that covered by the current checklist?

kusst:
What about very low FPS? I remember some CTimer::GetTimeStepFix-based physics hack was causing some bugs when game was running on very low FPS and it had to be reverted.

withmorten:
> 
> 
> What about very low FPS? I remember some CTimer::GetTimeStepFix-based physics hack was causing some bugs when game was running on very low FPS and it had to be reverted.

Yes, #37b6d491e7b9965dd344493f6220a4aca487901e.

ZLau92:
I built this to test. Answered my own question from above, looks like cars are still light as air to push on foot at high fps in this fork. Might be worth adding to the todo checklist?
But I'd also add that it feels like cars move much slower in this fork than they do in the mainline of re3 with the framelimiter on. 
i.e., max speed seems lower, and the engine makes high-rev noise much sooner.

Other than that, super cool. Especially where particles and some basic physics are concerned. Kind of surreal feeling since I've become so accustomed to things being severely broken at high fps.

SonofUgly:
Braking on bikes is now much slower at higher FPS (almost don't work at all at very high FPS).

Nopey:
I wrote a commit for "Fast Ctrl+B debug camera at high FPS" a week or two ago.

https://github.com/Nopey/re3/commit/3829e2da1a03570bad90853c3ac4251d26da7f82

There's also another change in `src/extras/debugmenu.cpp`, which makes the debug menu close when ESC is hit.

Veyrdite:
> Do the car physics related changes fix the issue where you can push a car very easily on foot at high fps? Is that covered by the current checklist?

Sorry not yet.  Possibly the ped<->veh collision code is not pushing the ped away enough and is instead always pushing the veh by X force impulse every frame?  I was hesitant to include timestep in collision impulse forces (because collisions are supposed to be momentary, eg 1 frame long, regardless of the framerate); but it might be called for in this case.  Possibly also ped<->ped collisions suffer this same problem (getting catapulted when you jump on peds at the right angle). 

Solving these problems to match vanilla FPS behaviors may take some creative thinking.

> > What about very low FPS? I remember some CTimer::GetTimeStepFix-based physics hack was causing some bugs when game was running on very low FPS and it had to be reverted.
> 
> Yes, [37b6d49](https://github.com/GTAmodding/re3/commit/37b6d491e7b9965dd344493f6220a4aca487901e).

That particular FIX_BUGS one is gone and I have completely reworked how CTimer:: is used in ProcessWheel().  Not sure if that jitter problem was caused by an actual error or if it's a fundamental instability (undamp) of this physics engine at low FPS; I'll have to look when I get some more time.   

N.B. I observed similar behaviors when nullifying the velocity-impulses (Fwd & Right) too much in that func, breaking the loop does seem to cause interesting problems.

> Braking on bikes is now much slower at higher FPS (almost don't work at all at very high FPS).

I'll take a look, thankyou SonofUgly.

Veyrdite:
> Kind of surreal feeling since I've become so accustomed to things being severely broken at high fps.

Surreal is exactly what I'm going for.  I'm now at the point where I get confused whether or not my F4 key is working because mashing it doesn't seem to have much effect.

Veyrdite:
> > Braking on bikes is now much slower at higher FPS (almost don't work at all at very high FPS).
> 
> I'll take a look, thankyou SonofUgly.

@SonofUgly fixed in 763337c25508e5c4eb91c87166a346a5141cd6af,  too many factors of time were being included and as a result standard (non-handbrake) braking was too weak on bikes at high FPS.

Veyrdite:
Boat physics is super-stuffed, the way Rockstar tried to make it timestep-independent (ie not require a fixed FPS) is totally wrong.  This is going to require quite a bit of mathematical tomfoorlery to fix.   

Take for example the following bundle of happiness:
```
if((GetModelIndex() != MI_SKIMMER || m_skimmerThingTimer == 0.0f) && !bSeparateTurnForce){
	// No idea what exactly is going on here besides drag in YZ
	float fx = Pow(pBoatHandling->vecTurnRes.x, CTimer::GetTimeStep());
	float fy = Pow(pBoatHandling->vecTurnRes.y, CTimer::GetTimeStep());
	float fz = Pow(pBoatHandling->vecTurnRes.z, CTimer::GetTimeStep());
	m_vecTurnSpeed = Multiply3x3(m_vecTurnSpeed, GetMatrix());	// invert - to local space
	// TODO: figure this out
	float magic = 1.0f/(1000.0f * SQR(m_vecTurnSpeed.x) + 1.0f) * fx;
	m_vecTurnSpeed.y *= fy;
	m_vecTurnSpeed.z *= fz;
	float forceUp = (magic - 1.0f) * m_vecTurnSpeed.x * m_fTurnMass;
	m_vecTurnSpeed = Multiply3x3(GetMatrix(), m_vecTurnSpeed);	// back to world
	CVector com = Multiply3x3(GetMatrix(), m_vecCentreOfMass);
	ApplyTurnForce(forceUp*GetUp(), com + GetForward());
}
```

That Pow(x, TimeStep) makes it look like might be trying to solve FPS-related problems, but things go totally wrong after that.  If their solution is correct then the sum of all ApplyTurnForce()'s after (eg) 1 second of gameplay should be the same regardless of the framerate.  According to my recreational spreadsheeting It's not, by a long shot (note the last column):

time period (seconds) | numsteps | **timestep (seconds)** | turnresx | turnspeedx | **fx** | **magic** | **forceUp** | **integral(forceup, time period)**
-- | -- | -- | -- | -- | -- | -- | -- | --
1 | 1 | 1 | 0.89 | 0.005 | 0.8900 | 0.8683 | -.000659 | -0.000659
1 | 10 | 0.1 | 0.89 | 0.005 | 0.9884 | 0.9643 | -.000178 | -0.001785
1 | 100 | 0.01 | 0.89 | 0.005 | 0.9988 | 0.9745 | -.000128 | -0.012763
1 | 1000 | 0.001 | 0.89 | 0.005 | 0.9999 | 0.9755 | -.000123 | -0.122520

(I've omitted some unimportant constants in the above equations and chosen some others from an actual boat in-game, but you get the point)

To fix boat physics you need to fix all boat-physics related problems, even small ones, because it's a "closed loop system".  Eg a problem with turn forces can make the boat sink too low into the water, increasing its drag and making it slower.  A problem with linear thrust forces can make it rise out of the water too much, ruining handling (the rudder needs to be in water).  I learned this when trying to fix vehicles: the problem with not being able to drive as fast around corners was because the rear wheels were not entering skid state which was because too much aerodynamic drag was being applied to the card which caused it to drive slower and straighter.

I'm going to have to replace their Pow(x, timestep) with something that is only correct at the official framelimited FPS, then concoct a new creative fix using CTimer::GetTimeStepFix() that covers all FPS cases.  This is a lot more work than I expected, I think I might go back to bed for now :P

yasinkaraaslan:
Can you also fix pathfinding at high fps? It makes some missions unplayable.

Nopey:
> Question: are all the debug menu code & stuff in the src/extras/ folder new (non-vanilla) stuff? I spot FIX_BUGS in your commit and wonder if the world has gone too far :)

I used a FIX_BUGS wrapper, as Nick007J's commit to the same function also checks FIX_BUGS.
https://github.com/GTAmodding/re3/commit/7a71a47f743b1496c069cf5c4f2765877aea412b

Both of these are probably unnecessary, as the entire function body is already behind an `IMPROVED_CAMERA` check.

Tomasak:
Noticed you didnt include cosmetic change on to do list - flashing objects on Hud, be it timer, police stars etc.
Great job so far

ZLau92:
Another thing that might be worth looking into
When you're using freecam with high fps while driving, it takes much longer for the camera to center back behind the vehicle after a turn than it does at 30ps.

Veyrdite:
With that latest siren fix: I'm getting close to the goal of having the game 100% playable at high FPS.  The last major issue that can't be worked around (ignored) is boat performance: some boat missions softlock during the cutscenes (slow boats) or can't be completed within the timer (slow boats).

Unfortunately the cheat-menu's goto-mission functionality is broken at the moment, segfaulting my game.  Time to do some more debug :)

Veyrdite:
> Can you also fix pathfinding at high fps? It makes some missions unplayable.

Sorry, what do you mean by this?  AI driver car pathfinding?

FWIW I've had problems with AI drivers being useless in racing missions whilst running with the framelimiter on.  I can't remember if things were better on console or not.

yasinkaraaslan:
I meant pathfinding for peds. They usually get stuck to a car or a wall. Most obvious example of this is "Guardian Angels" mission where dealers must reach Diaz to trigger a dialogue so the mission can continue. But one of the dealers would get stuck if the frame limiter is off and the mission wouldn't continue. This also happens in re3.

Veyrdite:
Thankyou, I'll give that a go once the mission-select feature in the debug menu is working again.

SonofUgly:
Perhaps a good way of testing this is by running 2 instances of the game side by side and using a controller to control both of them at the same time. Adding specific teleport points to the debug menu, and then spawning in cars, would make testing everything easier and more consistent too.
Tried it out and it seems like vehicle speed/acceleration is still slower at high fps: [Oceanic](https://files.catbox.moe/5z18zd.mp4), [ The Driver mission](https://files.catbox.moe/0uqdc3.mp4)
 (left is master, right is PR)

erorcun:
> Unfortunately the cheat-menu's goto-mission functionality is broken at the moment, segfaulting my game. Time to do some more debug.

I think that's because missions should be started in the city they're given from. Like Cortez's missions will crash if you're in second city.

I think it's not hard to guess you put so much effort into this PR, thank you =) As you know, it will take some time to review it.

erorcun:
I found 2 problems on 60 FPS:

- Initial acceleration of Admiral is slower then original acceleration on 30 FPS.
- I pressed backward/brake button when I was cruising in a boat, firstly it decelerated, and then stopped for a ~3-4 seconds before starting reversing. I tried this on vanilla, I think it's shorter on it.

Except that, I(and the team, we talked that at the Discord server) think GetLogicalFramesPassed should be in a macro and that macro should always be 1 when FIX_BUGS is not defined. That way we can clean up the code/remove many FIX_BUGS.

Looking forward to approve this PR. :)

Veyrdite:
Thanks erorcun.

**Car speed problems:** I'll take a look once I have some spare time.  Thankyou everyone for reporting, I have not been comparing against vanilla builds (only with/without the framelimiter enabled in my own builds and I've assumed that my changes have been mathematically == 1 when the framelimiter is enabled.  Clearly not true.

**Boat problems:** yes boat physics is still a complete mess.  I've fixed a few issues, but it's still not enough.

> Except that, I(and the team, we talked that at the Discord server) think GetLogicalFramesPassed should be in a macro and that macro should always be 1 when FIX_BUGS is not defined

That shouldn't affect me & this branch, I only ever modify code within #ifdef FIX_BUGS anyway.  You can do whatever you want to GetLogicalFramesPassed when FIX_BUGS is not defined.  On that note: If you find I've made mods outside of FIX_BUGS then please tell me!  I've tidied a few bits in places but not in ways that should affect output binaries.

src/vehicles/Vehicle.cpp
```diff
@@ -974,14 +997,9 @@ CVehicle::ProcessBikeWheel(CVector &wheelFwd, CVector &wheelRight, CVector &whee
...
 		right = -contactSpeedRight/wheelsOnGround;
-#ifdef FIX_BUGS
-		// contactSpeedRight is independent of framerate but right has timestep as a factor
-		// so we probably have to fix this
-		right *= CTimer::GetTimeStepFix();
-#endif
 
 		if(wheelStatus == WHEEL_STATUS_BURST){
-			float fwdspeed = Min(contactSpeedFwd, fBurstBikeSpeedMax);
+			float fwdspeed = Min(contactSpeedFwd, fBurstBikeSpeedMax);
```
Margen67, line 1002: whitespace

PR #1217, opened 2021-07-10
Fix debug script loader crashing missions

Nopey:
**NOTE:** I am unable to run re3 binaries I build at the moment, something something can't open HEAD.WAV.
If either Nick or riko20 could verify that these changes solve the issue, that'd be great.

* open_script is now exposed in Script.h; perhaps it should be namespaced
as a static method on CTheScripts? I'm unsure what is preferred.
* I've moved the joypad code out of open_script to prevent buttons held
  down at mission load time from changing the scriptToLoad.

### DIFF
--- a/src/control/Script.cpp
+++ b/src/control/Script.cpp
@@ -4390,7 +4390,11 @@ CTheScripts::SwitchToMission(int32 mission)
 	CTimer::Suspend();
 	int offset = CTheScripts::MultiScriptArray[mission];
 	CFileMgr::ChangeDir("\\");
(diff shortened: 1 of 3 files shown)


Nopey:
I verified my game integrity with steam, can now run re3.
Fixed a mistake I made with the paths, I'm currently playing the game with main_d.scm.
I'll report once I'm able to pick Misty up from Luigi's

Nopey:
All seems well on my end, I'm able to pick up Misty while using main_d.scm

PR #1218, opened 2021-07-10
switch out current ini_parser.hpp with pulzed/mINI

withmorten:
preserves comments, if newly generated has correct order compared to source code etc (no weird warnings on 64bit!) ...


src/core/re3.cpp
```diff
@@ -179,15 +179,18 @@ CustomFrontendOptionsPopulate(void)
...
 
-linb::ini cfg;
 bool ReadIniIfExists(const char *cat, const char *key, uint32 *out)
 {
-	std::string strval = cfg.get(cat, key, "\xBA");
+	std::string strval = cfg.get(cat).get(key);
 	const char *value = strval.c_str();
 	char *endPtr;
-	if (value && value[0] != '\xBA') {
+	if (value && value[0]) {
```
erorcun, line 193: this needs to be able to read empty strings, otherwise empty strings won't override their gta3.set counterparts, but we want them to do

PR #1220, opened 2021-07-11
Merge simple changes from miami into master

Philpax:
This PR improves maintainability by bringing over changes from `miami` that do not impact how the game functions. There are a few changes that might be a little controversial, but I figured they were like that in `miami` for a reason. Will happily remove any particular change for approval.


src/text/Messages.cpp
```diff
@@ -419,10 +420,9 @@ CMessages::InsertStringInString(wchar *str1, wchar *str2)
 	for (i = 0; i < total_size; ) {
 #ifdef MORE_LANGUAGES
 		if ((CFont::IsJapanese() && *_str1 == (0x8000 | '~') && *(_str1 + 1) == (0x8000 | 'a') && *(_str1 + 2) == (0x8000 | '~'))
-			|| (*_str1 == '~' && *(_str1 + 1) == 'a' && *(_str1 + 2) == '~'))
-		{
+			|| (*_str1 == '~' && *(_str1 + 1) == 'a' && *(_str1 + 2) == '~')) {
 #else
-			if (*_str1 == '~' && *(_str1 + 1) == 'a' && *(_str1 + 2) == '~') {
+		if (*_str1 == '~' && *(_str1 + 1) == 'a' && *(_str1 + 2) == '~') {
```
madebr, line 425: I think this indentation is incorrect.

src/rw/MemoryHeap.cpp
```diff
@@ -7,6 +7,8 @@
 #include "FileLoader.h"
 #include "MemoryHeap.h"
 
+// TODO(MIAMI)
+
 #ifdef USE_CUSTOM_ALLOCATOR
```
madebr, line 12:
```suggestion
#ifdef USE_CUSTOM_ALLOCATOR
```

src/rw/MemoryHeap.cpp
```diff
@@ -339,7 +341,7 @@ CMemoryHeap::TidyHeap(void)
 	}
 }
 
-// 
+// MIAMI: this is empty
```
madebr, line 344:
```suggestion
//
```

Sergeanur:
These are VC changes that belong to miami only. We would only take bug fixes from VC, and those should be wrapped into `#ifdef FIX_BUGS` too.

PR #1236, opened 2021-07-13
Fix NearPed count going negative

Nopey:
Deconstructing peds would attempt to remove themselves from their nearPeds, but would fail to remove themselves from any near peds' nearPeds list if the deconstructing ped was in the final position of nearPeds, and only decrement the near ped's numNearPeds.


erorcun:
Thanks.

PR #1240, opened 2021-07-14
Disable motion blur during cutscenes

Anarky:
This commit disable the motion blur during cutscenes - the characters animation seems too slow which results with ghosting - and restore its state afterwards.

### DIFF
--- a/src/animation/CutsceneMgr.cpp
+++ b/src/animation/CutsceneMgr.cpp
@@ -20,6 +20,7 @@
 #include "RpAnimBlend.h"
 #include "ModelIndices.h"
 #include "TempColModels.h"
+#include "postfx.h"
 
 const struct {
 	const char *szTrackName;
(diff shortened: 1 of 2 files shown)


erorcun:
can you try to add this as an 3rd option to Motion Blur setting on menu?

aap:
Why don't you just disable motion blur completely in your personal config if you don't like it? i don't see any reason this should be in the code.

Anarky:
@erorcun I can do that, but I discovered that unskippable cutscenes - for example when you deliver people to Marty Chonks - are not covered with this patch, do you know where they are rendered?

@aap motion blur works perfectly when I play, but here is what I see during cutscenes; if it's a rendering issue on my system then I will open a bug ticket

Sergeanur:
Yeah this looks like motion blur needs fixing.

aap:
I don't really see the problem. does it look different in vanilla?

Sergeanur:
Very much, yeah

PR #1243, opened 2021-07-15
Fix debug script loader crashing missions

Nopey:
See also: #1217 (re3)

This is a draft because I think `miami` branch's debug script loader was already functional entirely off the back of the `scriptFile` global being used properly.

I'd like to finish looking into pole dropping before returning to this


Nick007J:
I think this PR is basically in the branch now. Check if it can be closed.

PR #1254, opened 2021-07-24
Prevent Catalina's heli from despawning

NABN00B:
Fixes possible sequence break on The Exchange.


Sergeanur:
So, how do I reproduce this? Just get into the tunnel during the last mission?

NABN00B:
> 
> 
> So, how do I reproduce this? Just get into the tunnel during the last mission?

Exactly. The helicopter will despawn, and there's even a visual key for this as its radar blip gets removed. When you drive out, the helicopter will spawn above the picnic area and starts making its way to the dam again. This is a problem in itself because it can be in a weird place when the player actually triggers the fight with it.

The sequence break is that the newly spawned helicopter doesn't have any special properties so it can be blown up immediately. This immediately spawns Maria at her usual place and despawns all Colombians, essentially skipping all the fighting at the dam.

aap:
i think `if(i == HELI_CATALINA) continue;` inside the loop is cleaner

NABN00B:
We still don't know what causes the heli to fly East if you go into a tunnel after it had landed at the dam. Perhaps we should track down and fix that too before merging this.

Sergeanur:
Shouldn't your fix address specifically that?

NABN00B:
> 
> 
> Shouldn't your fix address specifically that?

Only indirectly.

PR #1257, opened 2021-07-27
Fix bug triggering gnNumTempPedList assertion

Nopey:
This solves the gnNumTempPedList assertion.
To prove this works, change gapTempPedList's length to 11, and visit the Triad's basketball court.

This change allows the near sectors to have more peds than the gap-list can contain by truncating the gap-list to the nearest `ARRAY_SIZE(m_nearPeds)` peds in the peds seen-so-far, rather than only sorting and truncating after all peds have been visited.

Additionally, the gnNumTempPedList assert previously was oversensitive, considering a full list after pushing to be an error condition. (It's only an error to try and push to a full list)

### DIFF
--- a/src/peds/Ped.cpp
+++ b/src/peds/Ped.cpp
@@ -392,8 +392,21 @@ CPed::BuildPedLists(void)
 					if (ped != this && !ped->bInVehicle) {
 						float dist = (ped->GetPosition() - GetPosition()).Magnitude2D();
 						if (nThreatReactionRangeMultiplier * 30.0f > dist) {
+#ifdef FIX_BUGS
+							static_assert( ARRAY_SIZE(m_nearPeds) < ARRAY_SIZE(gapTempPedList) - 1, "gapTempPedList needs wiggle room for unsorted peds and nil slot" );
+							// If the gap ped list is full, sort it and truncate it
+							// before pushing more unsorted peds
+							if( gnNumTempPedList == ARRAY_SIZE(gapTempPedList) - 1 )
+							{
+								gapTempPedList[gnNumTempPedList] = nil;
+								SortPeds(gapTempPedList, 0, gnNumTempPedList - 1);
+								gnNumTempPedList = ARRAY_SIZE(m_nearPeds);
+							}
+#endif
+
 							gapTempPedList[gnNumTempPedList] = ped;
 							gnNumTempPedList++;
+							// NOTE: We cannot absolutely fill the gap list, as the list is null-terminated before being passed to SortPeds
 							assert(gnNumTempPedList < ARRAY_SIZE(gapTempPedList));
 						}
 					}
--- a/src/peds/Ped.cpp
+++ b/src/peds/Ped.cpp
@@ -393,7 +393,6 @@ CPed::BuildPedLists(void)
 						float dist = (ped->GetPosition() - GetPosition()).Magnitude2D();
 						if (nThreatReactionRangeMultiplier * 30.0f > dist) {
 #ifdef FIX_BUGS
-							static_assert( ARRAY_SIZE(m_nearPeds) < ARRAY_SIZE(gapTempPedList) - 1, "gapTempPedList needs wiggle room for unsorted peds and nil slot" );
 							// If the gap ped list is full, sort it and truncate it
 							// before pushing more unsorted peds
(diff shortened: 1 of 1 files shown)


Nopey:
hang on, just caught a sneaky `gapTempPedList[gnNumTempPedList] = nil;` access, the assert was *not* overeager
gonna fix that and then it should be fine

Nopey:
Alright, the list is once again (properly?) null terminated, and ensures there's space for null termination.

src/peds/Ped.cpp
```diff
@@ -392,8 +392,21 @@ CPed::BuildPedLists(void)
...
 						if (nThreatReactionRangeMultiplier * 30.0f > dist) {
+#ifdef FIX_BUGS
+							static_assert( ARRAY_SIZE(m_nearPeds) < ARRAY_SIZE(gapTempPedList) - 1, "gapTempPedList needs wiggle room for unsorted peds and nil slot" );
+							// If the gap ped list is full, sort it and truncate it
+							// before pushing more unsorted peds
+							if( gnNumTempPedList == ARRAY_SIZE(gapTempPedList) - 1 )
+							{
+								gapTempPedList[gnNumTempPedList] = nil;
+								SortPeds(gapTempPedList, 0, gnNumTempPedList - 1);
+								gnNumTempPedList = ARRAY_SIZE(m_nearPeds);
```
erorcun, line 403: Why? `m_nearPeds` isn't used in this loop at all

src/peds/Ped.cpp
```diff
@@ -392,8 +392,21 @@ CPed::BuildPedLists(void)
 					if (ped != this && !ped->bInVehicle) {
 						float dist = (ped->GetPosition() - GetPosition()).Magnitude2D();
 						if (nThreatReactionRangeMultiplier * 30.0f > dist) {
+#ifdef FIX_BUGS
+							static_assert( ARRAY_SIZE(m_nearPeds) < ARRAY_SIZE(gapTempPedList) - 1, "gapTempPedList needs wiggle room for unsorted peds and nil slot" );
```
erorcun, line 396: This is a thing since C++11, so it's not allowed here, sorry.

src/peds/Ped.cpp
```diff
@@ -392,8 +392,21 @@ CPed::BuildPedLists(void)
...
+							// If the gap ped list is full, sort it and truncate it
+							// before pushing more unsorted peds
+							if( gnNumTempPedList == ARRAY_SIZE(gapTempPedList) - 1 )
```
erorcun, line 399: Instead of doing this, you can `break` all loops. It's clear that it can't continue further this point, because `gapTempPedList` will be full.

src/peds/Ped.cpp
```diff
@@ -392,8 +392,21 @@ CPed::BuildPedLists(void)
...
+							{
+								gapTempPedList[gnNumTempPedList] = nil;
+								SortPeds(gapTempPedList, 0, gnNumTempPedList - 1);
+								gnNumTempPedList = ARRAY_SIZE(m_nearPeds);
+							}
```
erorcun, line 404: Why don't you break here? It's clear that it can't continue anymore.

erorcun (commented):
In addition, this PR won't prevent triggering the assert, since you don't increase the array size or break the loops. Real reason of this crash was there being too much near peds to be cached.

Nopey:
The point of setting `gnNumTempPedList` to `ARRAY_SIZE(m_nearPeds)` is so that we can truncate the gap list to the nearest 10 peds.
Because we havn't visited all peds, we cannot simply `break`, as that will ignore peds that are potentially closer

Nopey:
By truncating it to the (so-far) closest 10, this frees up the other 40 slots of the gap list for unsorted (and potentially closer) peds

src/peds/Ped.cpp
Nopey:
```c
core\common.h
368:#define static_assert(bool_constexpr, message)
376:    static_assert(s == t, "Invalid structure size");
```
It's both defined and used in common.h

erorcun:
Oh I see what you did there. This is very clever, yeah. Lgtm if you change static_assert with something else. I'm aware that static_assert is used a few times in the project (outside CHECK_STRUCT_SIZES I mean), these are also wrong, and I hope will be removed at some point.

Sergeanur (approved):

PR #1258, opened 2021-07-29
Add define to enable/disable external 3d audio simulation

Sergeanur:
PS2 version doesn't use any 3d audio engine and the game just handles 3d positioning by itself. PC version lets MSS to handle 3d spatialization. It was now wrapped into define to be disabled for PS2 or just to play with PS2-esque audio experience. Can't be disabled on OpenAL without major surgery and IMO it ain't worth it since it only works in 3d space originally.


Sergeanur:
This needs to have m_nEmittingVolume sets into ifdef.

Issue #1180, opened 2021-06-30
[re3] undefined reference to `glfwGetX11Display`


bugaevc:
Commit https://github.com/GTAmodding/re3/commit/bf7280b55bffd568b00d5145f5799c7f73114ffc breaks building re3 on freedesktop systems that don't use X11 and/or don't build GLFW with `GLFW_EXPOSE_NATIVE_X11`.

Namely, this manifests as a linking error:

```
/usr/bin/ld: obj/linux-amd64-librw_gl3_glfw-oal/Debug/re3/glfw.o: in function `initkeymap()':
/re3/build/../src/skel/glfw/glfw.cpp:1473: undefined reference to `glfwGetX11Display'
/usr/bin/ld: obj/linux-amd64-librw_gl3_glfw-oal/Debug/re3/glfw.o: in function `checkKeyPresses()':
/re3/build/../src/skel/glfw/glfw.cpp:1687: undefined reference to `glfwGetX11Display'
collect2: error: ld returned 1 exit status
```

I have encountered this on several major GNU/Linux distributions:
* Fedora 34
* Ubuntu 20.04
* Arch Linux (latest at the moment of writing, using `glfw-wayland` package)

This is easily worked around by reverting the commit (which is not needed when not using X11 anyway).

Would be great if disabling this code was done automatically at configuration time.

erorcun:
You just need to disable `#define GET_KEYBOARD_INPUT_FROM_X11`

bugaevc:
I know nothing about Premake, so I cannot exactly help you with that, but say in Meson, you'd need to do something like this to detect whether `glfwGetX11Display()` exists, and only define `GET_KEYBOARD_INPUT_FROM_X11` in that case:

```meson
test_code = '''
#define GLFW_EXPOSE_NATIVE_X11
#include <GLFW/glfw3.h>
#include <GLFW/glfw3native.h>

void foo() { glfwGetX11Display(); }
'''

cxx = meson.get_compiler('cpp')
if cxx.links(test_code, dependencies: glfw)
    add_project_arguments('-DGET_KEYBOARD_INPUT_FROM_X11', language: 'cpp')
endif
```

And in CMake, like this:

```cmake
include(CheckSymbolExists)

set(CMAKE_REQUIRED_LIBRARIES glfw)
set(CMAKE_REQUIRED_DEFINITIONS "-DGLFW_EXPOSE_NATIVE_X11")
[...]

PR #1262, opened 2021-07-31
(re3) Fix compiler errors related to MinGW

alextwothousand:
* pragma comment is not supported with the mingw / msys2 compiler toolchain. I have mitigated this by adding a mingw compiler check in src/CMakeLists.txt
* You cannot cast from a HKL (HANDLE, PVOID) to a DWORD. I have mitigated this by casting to a uintptr first, then casting to a DWORD.


erorcun:
now it compiles on MinGW? :o

withmorten:
Perhaps casting to DWORD_PTR is better, since that's essentially the same?

src/CMakeLists.txt
```diff
@@ -29,6 +29,11 @@ target_link_libraries(${EXECUTABLE} PRIVATE
     Threads::Threads
 )
 
+if (MINGW)
+    target_link_libraries(${EXECUTABLE} PRIVATE
+            winmm)
```
madebr, line 34: I'm surprised MSVC doesn't need this.
What missing symbol is the linker complaining about?

madebr: Or is this using the msys2 compiler of msys2?

msys2 has (more then) 3 platforms: MSYS2, MINGW32 and MINGW64
- msys2 is a kind of unix on Windows
- mingw32 is pure 32-bit windows
- mingw64 is pure 64-bit windows

alextwothousand:
> 
> 
> now it compiles on MinGW? :o

always did, haha.

alextwothousand:
> 
> 
> Perhaps casting to DWORD_PTR is better, since that's essentially the same?

did actually try that. game wouldn't load, but I think it was for a unrelated issue. will test it again now and let you know how it goes.

src/CMakeLists.txt
alextwothousand: MSVC does need winmm. It still has access to it, I changed `#if _WIN32` to `#if _MSC_VER`.
I am using msys2 / mingw64.

If there is a `if(WIN32)` branch, I will likely move it there instead.

alextwothousand: I just looked into `src/CMakeLists.txt`. There is no `if(WIN32)` branch at all. Only a check for GNU / Clang OR MSVC compilers - which our situation does not match either.

alextwothousand:
I've made a few corrections.

src/CMakeLists.txt
madebr: Thanks!

src/skel/glfw/glfw.cpp
```diff
@@ -233,7 +233,9 @@ psGrabScreen(RwCamera *pCamera)
  *****************************************************************************
  */
 #ifdef _WIN32
+#ifdef _MSC_VER
```
madebr, line 236: This syntax is also used in `src/audio/oal/stream.cpp`.
```suggestion
#if defined _MSC_VER && !defined CMAKE_NO_AUTOLINK
```

alextwothousand: Grazie grazie.

erorcun:
I appreciate your fix for X11, but I want to do this by checking if environment has X11 (see https://github.com/GTAmodding/re3/issues/1180), and otherwise disable GET_KEYBOARD_INPUT_FROM_X11. if you remove that part, I can merge these 2 commits immediately.

madebr:
re3 needs to link to `X11` because it is calling X11 function directly:
https://github.com/GTAmodding/re3/blob/9a7fa478578beaba947ea867c15a25e411d641d8/src/skel/glfw/glfw.cpp#L1654

PR #1263, opened 2021-07-31
(reVC) Fix build errors related to MinGW

alextwothousand:
* pragma comment is not supported with the mingw / msys2 compiler toolchain. I have mitigated this by adding a mingw compiler check in src/CMakeLists.txt
* You cannot cast from a HKL (HANDLE, PVOID) to a DWORD. I have mitigated this by casting to a uintptr first, then casting to a DWORD.
* float.h (or cfloat) must be included regardless of what platform you are running - it is only not necessary with the msvc compiler. I have mitigated this by replacing the _WIN32 check with _MSC_VER, on line 17 (see #46eaa22).


withmorten:
You can just remove the ifdef check and include float.h always!

alextwothousand:
> 
> 
> You can just remove the ifdef check and include float.h always!

Potentially. Wasn't so sure how happy people would be with that.
Plus, I am more of an advocate for keeping header files inside header files.

Depending on what this project's styling guidelines are, I will make the necessary changes. Please inform me when you able to do so.

erorcun:
We prefer including header files in .cpp, whenever possible. You can add any header to any .cpp any time you want :)

alextwothousand:
> 
> 
> We prefer including header files in .cpp, whenever possible. You can add any header to any .cpp any time you want :)

Done.

PR #1271, opened 2021-08-02
Unify {ps,Rw}Timer type for all platforms

arkamar:
Both `tv_sec` and `tv_nsec` used in `psTimer` of non-windows systems are integers and there is no need to convert them to double.


erorcun:
No, converting tv_nsec to ms will result in double value, and this is beneficial for us.
In Timer.cpp you can see that we save fractional part (frameTimeFraction) and add it to timer when it reaches 1ms.

arkamar:
Ok, I see.

PR #1275, opened 2021-08-02
Experimental MPEG stuff

erorcun:


src/skel/win/win.cpp
```diff
@@ -948,11 +955,61 @@ void HandleGraphEvent(void)
...
+	ASSERT(pME != nil);
+
+	HRESULT hr = S_OK;
+
+	LONG evCode;
+	// We're 99% percent sure this was originally LONG as in Windows samples, so it was going to destroy the stack by trying to store 64 bit pointer into 32 bit.
+	LONG_PTR evParam1, evParam2;
+
+	// Process all queued events
+	while (SUCCEEDED(pME->GetEvent(&evCode, (LONG_PTR*)&evParam1, (LONG_PTR*)&evParam2, 0)))
```
withmorten, line 980: You can remove these casts

src/skel/win/win.cpp
```diff
@@ -948,11 +955,61 @@ void HandleGraphEvent(void)
...
+}
+#else
+void HandleGraphEvent(void)
+{
+	ASSERT(pME != nil);
+
+	HRESULT hr = S_OK;
+
+	LONG evCode;
+	// We're 99% percent sure this was originally LONG as in Windows samples, so it was going to destroy the stack by trying to store 64 bit pointer into 32 bit.
```
withmorten, line 976: I don't think this comment is necessary, since for 32bit builds LONG_PTR is LONG anyway and the resulting code is the same - and the type LONG_PTR is indicator enough it's only for 64bit.

PR #1282, opened 2021-08-07
High Framerate Improvements

ZLau92:
Improvements include
- In-vehicle freecam: angle limits, rotation speed, and fov
- droplet movement at high framerates
- The infamous bug where the player is thrown across the map for treading an uneven surface without a framelimiter is addressed
- Also corrects the issue where the player can push vehicles around as if they were feather-light


ZLau92:
-> Crash when pause / unpause while in vehicle
-> Review ped velocity - m_moved

PR #1284, opened 2021-08-08
Highfps

ZLau92:
Sorry for the pr spam.

Improvements include
- In-vehicle freecam: angle limits, rotation speed, and fov
- droplet movement at high framerates

Additionally, [this commit](https://github.com/GTAmodding/re3/commit/bb300a96b54e5995f73c6d63197f36d791004e8b) is to the following effect:
- The infamous bug where the player is thrown across the map for treading an uneven surface without a framelimiter no longer occurs
- Also corrects the issue where the player can push vehicles around as if they were feather-light, though I'd add that they seem a touch heavier at higher fps with this velocity adjustment.

^^ @Nick007J  We discussed this bit yesterday. What was the disposition on this? Does it look like I'm squaring GetTimeStepFix()? I think I understood that it's comparing animation progress to a ms timer, but not sure.
In either case, let me know if I should revert this bit so we can tackle this another way / time. (aap mentioned porting fixes from VC/SA in the case of flying off of ledges)


ZLau92:
reverted that physics timing manipulation in my High FPS pr. In testing, I found that it had some unintended effects on climbing some sets of stairs when testing in vc.

